PR #111, opened 2024-08-07
FMWK-530 Fix handlers .Wait() context

filkeith:
- removed context from handlers .Wait() methods
- pass global backup context to handlers

### DIFF
--- a/client.go
+++ b/client.go
@@ -167,7 +167,7 @@ func (c *Client) Backup(
 		return nil, err
 	}
 
-	handler := newBackupHandler(config, c.aerospikeClient, c.logger, writer)
+	handler := newBackupHandler(ctx, config, c.aerospikeClient, c.logger, writer)
 	handler.run(ctx)
 
 	return handler, nil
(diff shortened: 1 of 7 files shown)


client.go
```diff
@@ -167,7 +167,7 @@ func (c *Client) Backup(
 		return nil, err
 	}
 
-	handler := newBackupHandler(config, c.aerospikeClient, c.logger, writer)
+	handler := newBackupHandler(ctx, config, c.aerospikeClient, c.logger, writer)
```
korotkov-aerospike, line 170: We pass same ctx to same struct twice.
I think, we should only pass it to one or another.

korotkov-aerospike (approved):

client.go
filkeith: https://aerospike.atlassian.net/browse/FMWK-531

reugn (approved):
As to our discussion, we'll need to refactor the logic and evaluate the option of returning the context parameter.

PR #113, opened 2024-08-11
APPS-1241 Add concurrent scans limiter

korotkov-aerospike:


io/aerospike/record_reader_test.go
```diff
@@ -70,6 +71,7 @@ func (suite *readersTestSuite) TestAerospikeRecordReader() {
 	)
 
 	reader := NewRecordReader(
+		context.TODO(),
```
filkeith, line 74: If it is test why not to use context.Background(). As context.TODO() is used as TODO for future, when we don't know what context to use right now, and plan to replace it in future. But it is test, we don't have such problem here

client.go
```diff
@@ -90,33 +91,60 @@ type AerospikeClient interface {
 type Client struct {
 	aerospikeClient AerospikeClient
 	logger          *slog.Logger
+	scanLimiter     *semaphore.Weighted
 	id              string
 }
 
+type Option func(*Client)
```
reugn, line 98:
```suggestion
// Option is a functional option that allows configuring the [Client].
type Option func(*Client)
```

client.go
```diff
@@ -90,33 +91,60 @@ type AerospikeClient interface {
...
+
+// WithID sets the ID for the Client.
```
reugn, line 100:
```suggestion
// WithID sets the ID for the [Client].
```

client.go
```diff
@@ -90,33 +91,60 @@ type AerospikeClient interface {
...
+func WithID(id string) Option {
+	return func(c *Client) {
+		c.id = id
+	}
+}
+
+// WithLogger sets the logger for the Client.
```
reugn, line 107:
```suggestion
// WithLogger sets the logger for the [Client].
```

client.go
```diff
@@ -90,33 +91,60 @@ type AerospikeClient interface {
...
+func WithLogger(logger *slog.Logger) Option {
+	return func(c *Client) {
+		c.logger = logger
+	}
+}
+
+// WithScanLimiter sets the scanLimiter for the Client.
```
reugn, line 114:
```suggestion
// WithScanLimiter sets the scan limiter for the [Client].
```

client.go
```diff
@@ -90,33 +91,60 @@ type AerospikeClient interface {
...
+	return func(c *Client) {
+		c.scanLimiter = sem
+	}
+}
+
 // NewClient creates a new backup client.
+// Options:
 //   - ac is the aerospike client to use for backup and restore operations.
 //   - id is an identifier for the client.
 //   - logger is the logger that this client will log to.
```
reugn, line 125:
```suggestion
// NewClient creates a new backup client.
//   - ac is the aerospike client to use for backup and restore operations.
//
// Options:
//   - [WithID] to set an identifier for the client.
//   - [WithLogger] to set a logger that this client will log to.
//   - [WithScanLimiter] to set a semaphore that is used to limit number of
//     concurrent scans.
```

client.go
reugn, line 98: It's better to name it `ClientOption` or `ClientOpt`.

examples/aws/s3/main.go
```diff
@@ -17,12 +17,10 @@ package main
 import (
 	"context"
 	"fmt"
-	"log"
-	"log/slog"
-
 	"github.com/aerospike/aerospike-client-go/v7"
 	"github.com/aerospike/backup-go"
 	"github.com/aerospike/backup-go/io/aws/s3"
+	"log"
```
reugn, line 23: Format imports.

examples/readme/main.go
```diff
@@ -16,11 +16,9 @@ package main
 
 import (
 	"context"
-	"log"
-	"log/slog"
-
 	"github.com/aerospike/aerospike-client-go/v7"
 	"github.com/aerospike/backup-go"
+	"log"
```
reugn, line 21: Format imports.

reugn (commented):
* Add coverage for the client options.

reugn (approved):

PR #114, opened 2024-08-15
Add retry mechanism to restore writes

korotkov-aerospike:


korotkov-aerospike:
make sense to use same RetryPolicy object on backup

io/aerospike/record_batch_writer.go
```diff
@@ -79,25 +81,58 @@ func (rw *batchRecordWriter) flushBuffer() error {
...
+	var attempt int
+	for attemptsLeft(rw.retryPolicy, attempt) {
+		err := rw.asc.BatchOperate(nil, rw.operationBuffer)
+
+		if isNilOrAcceptableError(err) {
+			rw.operationBuffer = rw.processAndFilterOperations()
+			if len(rw.operationBuffer) == 0 {
+				return nil // All operations succeeded
+			}
+		} else if !shouldRetry(err) {
```
filkeith, line 93: May be avoid `else if ` if not necessary. Like:

```go
if !shouldRetry(err) {
    return fmt.Errorf("non-retryable error on restore: %w", err)
}
```

then

```go
if isNilOrAcceptableError(err) {
    rw.operationBuffer = rw.processAndFilterOperations()
        if len(rw.operationBuffer) == 0 {
             return nil // All operations succeeded
         }
}
```

io/aerospike/record_writer.go
```diff
@@ -26,6 +26,7 @@ type singleRecordWriter struct {
 	asc         dbWriter
 	writePolicy *a.WritePolicy
 	stats       *models.RestoreStats
+	retryPolicy *models.RetryPolicy
```
filkeith, line 29: May be I missing smth, but didn't *a.WritePolicy.MaxRetry do the same? or we want to retry only on certain errors?

filkeith (changes requested):

io/aerospike/record_batch_writer.go
korotkov-aerospike: unfortunately, it's not that simple, acceptable errors should not be retried.

io/aerospike/record_writer.go
korotkov-aerospike: yes, not all errors are retryable. But more important, we want to avoid having too much connections to aerospike, so when we do our own retry mechanism, we don't hold connection while sleeping.

filkeith (approved):

reugn (approved):

PR #120, opened 2024-08-19
FMWK-538 Add GCP storage support

filkeith:
- added gcp storage support
- disabled `fieldalignment` linter, as it makes difficult to write clean code


io/gcp/storage/reader_writer_test.go
```diff
@@ -0,0 +1,496 @@
...
+type GCPSuite struct {
+	suite.Suite
+	client *storage.Client
+}
+
+func (s *GCPSuite) SetupSuite() {
+	fmt.Println("setting up suite")
+	ctx := context.Background()
+	client, err := storage.NewClient(ctx, option.WithEndpoint(testServiceAddress), option.WithoutAuthentication())
+	if err != nil {
```
korotkov-aerospike, line 59: we can use Require.noError here too

io/gcp/storage/reader.go
```diff
@@ -0,0 +1,141 @@
...
+	folderName string,
+	validator validator,
+) (*StreamingReader, error) {
+	if validator == nil {
+		return nil, fmt.Errorf("validator cannot be nil")
+	}
+
+	prefix := folderName
+	// Protection from incorrect input.
+	if !strings.HasSuffix(folderName, "/") && folderName != "/" && folderName != "" {
```
korotkov-aerospike, line 60: I'm not fully understand this protection, you sure about it? If the folder name is empty string, then prefix will be mpty too... is it ok? Maybe return error instead?

korotkov-aerospike (approved):

io/gcp/storage/reader.go
filkeith: if folder is empty, than it will be root. I just want to protect from incorrect input. So user can input
"/"
""
"folder"
"folder/"

reugn (approved):

PR #121, opened 2024-08-21
FMWK-529 Extract IO operations from configuration structs

filkeith:


config_secret_agent.go
```diff
@@ -46,108 +38,28 @@ type SecretAgentConfig struct {
...
-	if err != nil {
-		return "", fmt.Errorf("failed to initialize secret agent client: %w", err)
+	if s.ConnectionType == nil {
+		return fmt.Errorf("connection type is required")
 	}
 
-	result, err := saClient.GetSecret(resource, secretKey)
-	if err != nil {
-		return "", fmt.Errorf("failed to get secret from secret agent: %w", err)
+	cType := *s.ConnectionType
```
reugn, line 59: Can be `nil`.

secret_agent.go
```diff
@@ -0,0 +1,123 @@
...
+	"os"
+	"strings"
+	"time"
+
+	saClient "github.com/aerospike/backup-go/pkg/secret-agent"
+)
+
+const secretPrefix = "secrets:"
+
+func getSecret(config *SecretAgentConfig, key string) (string, error) {
```
reugn, line 30: Worth checking config for nil.

config_secret_agent.go
filkeith: if it will be nil, it will return error 3 lines upper
https://github.com/aerospike/backup-go/blob/FMWK-529-move-logic-from-config/config_secret_agent.go#L55

secret_agent.go
filkeith: there is nil check here
https://github.com/aerospike/backup-go/blob/FMWK-529-move-logic-from-config/io_encryption.go#L108

but yes, I'll move it inside `getSecret()`

config_secret_agent.go
filkeith: changed if

secret_agent.go
```diff
@@ -0,0 +1,126 @@
...
+
+	result, err := client.GetSecret(resource, secretKey)
+	if err != nil {
+		return "", fmt.Errorf("failed to get secret from secret agent: %w", err)
+	}
+
+	return result, nil
+}
+
+func getResourceKey(key string) (resource, secretKey string, err error) {
```
korotkov-aerospike, line 83: Maybe return one object here?
anyway you later merge them
	msg := models.Request{
		Resource:  resource,
		SecretKey: secretKey,
	}

filkeith: didn't want to create new entity for one function call. Also this place wont change (like no other fields will be added), as it is used for requesting secret agent. I thin thank named return will be enough

reugn (approved):

PR #122, opened 2024-08-22
FMWK-531 Revise handler's context 

filkeith:
- removed unnecessary context  from handler.run()
- added cancel function to global handler contrext
- added new context to handler.run.Wait(ctx), so if we interrupt .Wait() func we will cancel global context and stop backup

### DIFF
--- a/client.go
+++ b/client.go
@@ -156,32 +156,32 @@ func NewClient(ac AerospikeClient, opts ...ClientOpt) (*Client, error) {
 	return client, nil
 }
 
-func (c *Client) getUsableInfoPolicy(p *a.InfoPolicy) a.InfoPolicy {
+func (c *Client) getUsableInfoPolicy(p *a.InfoPolicy) *a.InfoPolicy {
 	if p == nil {
 		p = c.aerospikeClient.GetDefaultInfoPolicy()
 	}
 
-	return *p
+	return p
 }
 
-func (c *Client) getUsableWritePolicy(p *a.WritePolicy) a.WritePolicy {
+func (c *Client) getUsableWritePolicy(p *a.WritePolicy) *a.WritePolicy {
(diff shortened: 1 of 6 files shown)


filkeith:
As we discussed with @reugn, I've added different context to .Wait() method, so we can interrupt execution from different places

handler_restore.go
```diff
@@ -46,15 +46,19 @@ type StreamingReader interface {
 // RestoreHandler handles a restore job using the given reader.
 type RestoreHandler struct {
 	// Global backup context for a whole restore process.
-	ctx             context.Context
+	ctx       context.Context
+	ctxCancel context.CancelFunc
```
korotkov-aerospike, line 50: I think it's better call it cancelFunc, or just cancel

korotkov-aerospike (approved):

client.go
```diff
@@ -156,28 +156,28 @@ func NewClient(ac AerospikeClient, opts ...ClientOpt) (*Client, error) {
 	return client, nil
 }
 
-func (c *Client) getUsableInfoPolicy(p *a.InfoPolicy) a.InfoPolicy {
+func (c *Client) getUsableInfoPolicy(p *a.InfoPolicy) *a.InfoPolicy {
```
reugn, line 159: It is safer to return copies of the default policy structs here so that if we decide to customize them downstream, the defaults will stay unchanged.

handler_restore.go
filkeith: ok

client.go
filkeith: But we get default policy only if our custom policy is not set. I think that if we want to modify policy, we should modify custom policy. But ok, I'll add copy

reugn (approved):

PR #127, opened 2024-08-28
FMWK-393 Add extra-ttl flag support

filkeith:
- added extra-ttl support


config_restore.go
```diff
@@ -68,6 +68,8 @@ type RestoreConfig struct {
 	BatchSize int
 	// Max number of parallel writers to target AS cluster.
 	MaxAsyncBatches int
+	// Amount of extra time-to-live to add to records that have expirable void-times.
```
reugn, line 71: Worth documenting that the value is expected to be in seconds.

reugn (approved):

PR #129, opened 2024-08-29
Fix RecordReader: check for nil before release of semaphore

korotkov-aerospike:


korotkov-aerospike:
It was possible to get NPE if you try to close reader right after creation.

io/aerospike/record_reader.go
```diff
@@ -134,12 +134,12 @@ func (r *RecordReader) Read() (*models.Token, error) {
...
 // if it was started.
 func (r *RecordReader) Close() {
-	if r.config.scanLimiter != nil {
-		r.config.scanLimiter.Release(int64(len(r.scanResult.data)))
-	}
-
 	if r.scanResult != nil {
 		r.scanResult.Close()
+
+		if r.config.scanLimiter != nil {
```
reugn, line 140: @korotkov-aerospike, we need to release exactly the same weight we acquired (do not rely on the result). Otherwise, it may quickly exhaust the semaphore.
```go
if r.config.scanLimiter != nil {
    acquired := max(1, len(r.config.setList))
    r.config.scanLimiter.Release(int64(acquired))
}
```

reugn (approved):

PR #135, opened 2024-09-03
FWK-542 Add Azure Blob support

filkeith:


io/aws/s3/reader.go
```diff
@@ -210,7 +209,7 @@ func (r *Reader) streamFile(
 		Key:    &filename,
 	})
 	if err != nil {
-		errorsCh <- fmt.Errorf("failed to open %s: %w", filename, err)
+		errorsCh <- fmt.Errorf("failerd to create reader from file %s: %w", filename, err)
```
reugn, line 212:
```suggestion
		errorsCh <- fmt.Errorf("failed to create reader from file %s: %w", filename, err)
```

io/azure/blob/writer.go
```diff
@@ -0,0 +1,243 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = <-w.done
+	if err != nil {
+		return err
+	}
+
+	return nil
```
reugn, line 185:
```suggestion
	return <-w.done
```

io/azure/blob/writer.go
```diff
@@ -0,0 +1,243 @@
...
+
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blob"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blockblob"
+)
+
+const (
+	uploadStreamFileType    = "application/octet-stream"
+	uploadStreamBlockSize   = 5 * 1024 * 1024 // 5MB, minimum size of a part
+	uploadStreamConcurrency = 5
```
reugn, line 32: It's better to make it an `option` and use this value by default.

io/azure/blob/writer.go
```diff
@@ -0,0 +1,243 @@
...
+
+func newBlobWriter(ctx context.Context, blobClient *blockblob.Client) io.WriteCloser {
+	pipeReader, pipeWriter := io.Pipe()
+
+	w := &blobWriter{
+		blobClient: blobClient,
+		ctx:        ctx,
+		pipeReader: pipeReader,
+		pipeWriter: pipeWriter,
+		done:       make(chan error),
```
reugn, line 148:
```suggestion
		done:       make(chan error, 1),
```

io/azure/blob/writer.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+
+	go w.uploadStream()
+
+	return w
+}
+
+func (w *blobWriter) uploadStream() {
+	defer w.pipeReader.Close()
```
reugn, line 157: Unhandled error on Close.

io/azure/blob/writer.go
```diff
@@ -0,0 +1,243 @@
...
+func (w *Writer) NewWriter(ctx context.Context, filename string) (io.WriteCloser, error) {
+	// protection for single file backup.
+	if !w.isDir {
+		if !w.called.CompareAndSwap(false, true) {
+			return nil, fmt.Errorf("parallel running for single file is not allowed")
+		}
+	}
+	// If we use backup to single file, we overwrite the file name.
+	if !w.isDir {
+		filename = w.path
```
reugn, line 120: Can be merged with the previous if.

io/azure/blob/writer.go
filkeith: it will be closed by backup, so removed this line

io/azure/blob/writer.go
```diff
@@ -154,12 +162,10 @@ func newBlobWriter(ctx context.Context, blobClient *blockblob.Client) io.WriteCl
 }
 
 func (w *blobWriter) uploadStream() {
-	defer w.pipeReader.Close()
-
 	contentType := uploadStreamFileType
 	_, err := w.blobClient.UploadStream(w.ctx, w.pipeReader, &azblob.UploadStreamOptions{
 		BlockSize:   uploadStreamBlockSize,
-		Concurrency: uploadStreamConcurrency,
+		Concurrency: uploadStreamConcurrencyDefault,
```
reugn, line 168: Take the value from options.

filkeith: fixed

reugn (approved):

PR #137, opened 2024-09-05
Restore process refactoring

korotkov-aerospike:
There are many files changed, but the PR itself is not very large. Here’s what was done:

* Split the parallelism configuration into separate Read and Write. One controls the number of AS scans, and the other controls the number of file writers.

* Moved the ProcessorWorker struct to the pipeline package, where it belongs, alongside ReadWorker and WriteWorker.

* Modified the token reader (used in restore) to accept a channel of readers. Instead of reading in batches, each token reader now retrieves a reader from the channel, processes it entirely, and then moves on to the next reader. This change allows us to process all backup files in a manner similar to a thread pool.

* Updated the Record field in the Token struct to be a pointer, to be consistent with *UDF and *SIndex.

* Introduced ComposedProcessor that composes multiple processors to reduce usage of channels.


pipeline/write_worker.go
```diff
@@ -87,6 +91,9 @@ func (w *writeWorker[T]) Run(ctx context.Context) (err error) {
 					return err
 				}
 			}
+		default:
+			emptyChannel.Add(1)
+			time.Sleep(10 * time.Millisecond)
```
reugn, line 96: Revert this debug code.

reugn (approved):

PR #146, opened 2024-09-17
FMWK-399 Add support for remove artifacts flag

filkeith:
- added remove-artifacts flag
- made RemoveFiles() public, so backup service can use it, added recursive folder iteration

### DIFF
--- a/cmd/asbackup/main.go
+++ b/cmd/asbackup/main.go
@@ -176,8 +176,13 @@ func run(cmd *cobra.Command, _ []string) error {
 	if err != nil {
 		return err
 	}
-	// Run app.
-	return asb.Run(cmd.Context())
+	// If RemoveArtifacts is set to true, we shall do it on ASBackup initialization.
+	// So asb will be nil, as we shouldn't do the backup, only remove artifacts.
+	if asb != nil {
+		return asb.Run(cmd.Context())
+	}
+
+	return nil
 }
 
 func printVersion() {
--- a/cmd/asbackup/main.go
+++ b/cmd/asbackup/main.go
@@ -176,13 +176,8 @@ func run(cmd *cobra.Command, _ []string) error {
 	if err != nil {
 		return err
 	}
-	// If RemoveArtifacts is set to true, we shall do it on ASBackup initialization.
-	// So asb will be nil, as we shouldn't do the backup, only remove artifacts.
-	if asb != nil {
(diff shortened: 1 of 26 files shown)


cmd/asbackup/main.go
```diff
@@ -176,8 +176,13 @@ func run(cmd *cobra.Command, _ []string) error {
 	if err != nil {
 		return err
 	}
-	// Run app.
-	return asb.Run(cmd.Context())
+	// If RemoveArtifacts is set to true, we shall do it on ASBackup initialization.
+	// So asb will be nil, as we shouldn't do the backup, only remove artifacts.
```
filkeith, line 180: or may be we should not check it in .Run() and do nothing

filkeith:
I'm thinking to rename .Validator to .Filter, as we are filtering file names

cmd/internal/app/writers.go
```diff
@@ -38,11 +38,13 @@ func newLocalWriter(b *models.Backup, c *models.Common) (backup.Writer, error) {
 		opts = append(opts, local.WithFile(b.OutputFile))
 	}
 
-	if b.RemoveFiles {
+	if b.RemoveFiles || b.RemoveArtifacts {
```
reugn, line 41: It would be better to add a method like `ShouldClearTarget` in models.Backup to make it less error prone.

io/gcp/storage/writer.go
```diff
@@ -144,21 +157,35 @@ func isEmptyDirectory(ctx context.Context, bucketHandle *storage.BucketHandle, p
...
 
 		// Skip files in folders.
-		if isDirectory(prefix, objAttrs.Name) {
+		if isDirectory(w.path, objAttrs.Name) && !w.withNestedDir {
 			continue
 		}
 
-		return false, nil
+		// If validator is set, remove only valid files.
+		if w.validator != nil {
```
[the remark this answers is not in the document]
filkeith: as we assume that user can create his own validators, i prefer to leave it like this. Because it will be not obvious for user to add nil check in the Run method.

cmd/internal/app/writers.go
filkeith: you mean read `--remove-files` and `--remove-artifacts` to one struct field `ShouldClearTarget` ?

reugn (approved):

PR #147, opened 2024-09-22
FMWK-563 Make check for emptiness only for directory

filkeith:
- we check only dir if it is empty

### DIFF
--- a/tests/integration/s3_integration_test.go
+++ b/tests/integration/s3_integration_test.go
@@ -223,15 +223,6 @@ func (s *writeReadTestSuite) writeSingleFile(filename string, bytes, times int, 
 		s.FailNow("failed to close writer", err)
 	}
 
-	// cannot create new streamingReader because folder is not empty
-	_, err = s3Storasge.NewWriter(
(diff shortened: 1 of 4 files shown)


tests/integration/s3_integration_test.go
```diff
@@ -223,15 +223,6 @@ func (s *writeReadTestSuite) writeSingleFile(filename string, bytes, times int,
 		s.FailNow("failed to close writer", err)
 	}
 
-	// cannot create new streamingReader because folder is not empty
```
korotkov-aerospike, line 226: well, it's a good test, we can leave it. 
just use WithDir


btw, there is misprint in s3Storasge

korotkov-aerospike: or, change assertion to "require no error"

filkeith: it is `writeSingleFile` test, so can't use `WithDir` here. Also with no error is is tested above

korotkov-aerospike (approved):

PR #148, opened 2024-09-22
FMWK-558 Add support for the compact flag

filkeith:
I checked asrestore works as is with not encoded data. SO no need to change it.


io/encoding/asb/encode.go
```diff
@@ -525,3 +567,9 @@ func udfToASB(udf *models.UDF, w io.Writer) (int, error) {
 		udf.Content,
 	)
 }
+
+// makeCompactPrefix adds ! to prefix, to show decoder that data in this string is not encoded with base64.
+// We don't do any safety check here to make it fast.
+func makeCompactPrefix(p []byte) []byte {
```
filkeith, line 573: May be I should not append, as it is slow operation. And add `binHLLCompactTypePrefix` variable with []byte("- Y! ") as it will be faster

reugn (approved):

PR #149, opened 2024-09-23
FMWK-405 Add support for node-list option

filkeith:
- added --node-list support

### DIFF
--- a/config_backup.go
+++ b/config_backup.go
@@ -42,6 +42,13 @@ type BackupConfig struct {
 	SecretAgentConfig *SecretAgentConfig
 	// Namespace is the Aerospike namespace to back up.
 	Namespace string
+	// NodeList contains a list of nodes to back up.
+	// <IP addr 1>:<port 1>[,<IP addr 2>:<port 2>[,...]]
+	// <IP addr 1>:<TLS_NAME 1>:<port 1>[,<IP addr 2>:<TLS_NAME 2>:<port 2>[,...]]
+	// Backup the given cluster nodes only.
+	// If it is set, ParallelNodes automatically set to true.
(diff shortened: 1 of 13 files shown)


shared.go
```diff
@@ -106,3 +106,38 @@ func splitNodes(nodes []*a.Node, numWorkers int) ([][]*a.Node, error) {
...
+		_, ok := nodesMap[nodeStr]
+		if ok {
+			filteredNodes = append(filteredNodes, nodes[i])
+		}
+	}
+
+	return filteredNodes
+}
+
+func nodeToString(node *a.Node) string {
```
filkeith, line 136: Built-in node.String() method don't add TLSName to the result, so I have to write my own.

config_backup.go
```diff
@@ -153,17 +160,21 @@ func (c *BackupConfig) validate() error {
...
-	if !c.ParallelNodes {
+	if !c.ParallelNodes && len(c.NodeList) == 0 {
 		if err := c.Partitions.validate(); err != nil {
 			return err
 		}
 	}
 
 	if c.AfterDigest != "" {
+		if c.ParallelNodes || len(c.NodeList) != 0 {
+			return fmt.Errorf("parallel by nodes/node liost and after digest at the same time not allowed")
```
reugn, line 175: liost -> list

shared.go
```diff
@@ -106,3 +106,38 @@ func splitNodes(nodes []*a.Node, numWorkers int) ([][]*a.Node, error) {
 
 	return result, nil
 }
+
+// filterNodes iterates on nodes and choose only nods that are in nodesList.
```
reugn, line 110:
```suggestion
// filterNodes iterates over the nodes and selects only those nodes that are in nodesList.
```

reugn (approved):

PR #152, opened 2024-09-23
FMWK-564 Add support for the prefer-racks flag

filkeith:
- added prefer racks
- fixed error with nodes and partitions backup

### DIFF
--- a/cmd/internal/app/clients.go
+++ b/cmd/internal/app/clients.go
@@ -17,6 +17,7 @@ package app
 import (
 	"context"
 	"fmt"
+	"strconv"
 
 	"cloud.google.com/go/storage"
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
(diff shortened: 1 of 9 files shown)


cmd/internal/app/clients.go
```diff
@@ -135,3 +148,27 @@ func toHosts(htpSlice client.HostTLSPortSlice) []*aerospike.Host {
 
 	return hosts
 }
+
+func parseRacks(racks string) ([]int, error) {
+	racksStringSlice := stringSplit(racks)
+	racksIntSlice := make([]int, len(racksStringSlice))
```
reugn, line 154:
```suggestion
	racksIntSlice := make([]int, 0, len(racksStringSlice))
```

cmd/internal/app/configs.go
```diff
@@ -275,3 +282,11 @@ func mapRetryPolicy(r *models.Restore) *bModels.RetryPolicy {
 		MaxRetries:  r.RetryMaxRetries,
 	}
 }
+
+func stringSplit(s string) []string {
```
reugn, line 286: It's better to give it a more descriptive name like `splitByComma`.

reugn (approved):

PR #158, opened 2024-10-07
FMWK-570 Add backup state/continuation for backup/restore operation

filkeith:
How it works:
1. We must set FileLimit, StateFile name and PageSize (it is used for paginated reading from DB. Less PageSize less results overlap, and speed)
2. We read data from db page by page. And save page state fro each record.
3. Workers read->process->write in this mode work in "straight mode", which mean that we initialize same number of all workers = number of partition groups. Each worker transfer data to next worker 1:1.
4. On encoding operation (right before writing) we save current page to memory.
5. On Writer.Close() when we reached FileLimit we save current page state for this worker to disk.
6. On resume/continue we read from file current pages and start to read them from the beginning.
7. So result will be overlapped little bit.


tests/integration/integration_test.go
```diff
@@ -1103,6 +1112,128 @@ func (suite *backupRestoreTestSuite) TestBackupEstimateOk() {
...
+	bFolder := fmt.Sprintf("%s_%d", testFolder, i)
+
+	writers, err := local.NewWriter(
+		ctx,
+		local.WithValidator(asb.NewValidator()),
+		local.WithSkipDirCheck(),
+		local.WithDir(bFolder),
+	)
+	if err != nil {
+		panic(err)
```
reugn, line 1156: It is better to avoid panicking. Use something like `suite.FailNow(err.Error())`.

writers.go
```diff
@@ -85,21 +85,32 @@ func (tw *tokenStatsWriter) Close() error {
...
 type tokenWriter struct {
-	encoder Encoder
-	output  io.Writer
-	logger  *slog.Logger
+	encoder          Encoder
+	output           io.Writer
+	logger           *slog.Logger
+	recordsStateChan chan<- models.PartitionFilterSerialized
+	// Number of writer.
+	n int
```
reugn, line 93: Consider wrapping those two into a state related structure, since they are tightly coupled (something like `StateInfo`).

state.go
```diff
@@ -0,0 +1,271 @@
...
+
+func (s *State) serveRecords() {
+	var counter int
+
+	for {
+		select {
+		case <-s.ctx.Done():
+			return
+		case state := <-s.RecordsStateChan:
+			if state.Begin == 0 && state.Count == 0 && state.Digest == nil {
```
reugn, line 234: Consider moving this check into a method with a meaningful name.

state.go
```diff
@@ -0,0 +1,271 @@
...
+				continue
+			}
+
+			counter++
```
reugn, line 238: The `counter` variable is not read.

state.go
```diff
@@ -0,0 +1,271 @@
...
+
+	if n > -1 {
+		s.RecordStatesSaved[n] = s.RecordStates[n]
+	}
+
+	if err = enc.Encode(s); err != nil {
+		return fmt.Errorf("failed to encode state data: %w", err)
+	}
+
+	file.Close()
```
reugn, line 182: Log the error.

state.go
```diff
@@ -0,0 +1,271 @@
...
+	// Run watcher on initialization.
+	go s.serve()
+	go s.serveRecords()
+
+	return &s, nil
+}
+
+// serve dumps files to disk.
+func (s *State) serve() {
+	for msg := range s.SaveCommandChan {
```
reugn, line 156: * It's better to add the global context awareness.
* `s.SaveCommandChan` is never closed.

io/aerospike/record_reader_paginated.go
```diff
@@ -0,0 +1,192 @@
...
+			// result contains []*a.Result and serialized filter models.PartitionFilterSerialized
+			result := make([]*pageRecord, 0, r.config.pageSize)
+
+			// to count records on pageRecord.
+			var counter int64
+			for res := range recSet.Results() {
+				counter++
+
+				if res.Err != nil {
+					continue
```
reugn, line 170: Consider logging the error.

io/aerospike/record_reader_paginated.go
```diff
@@ -0,0 +1,192 @@
...
+				result = append(result, newPageRecord(res, &curFilter))
+			}
+
+			if aErr = recSet.Close(); aErr != nil {
+				errChan <- fmt.Errorf("failed to close record set: %w", aErr.Unwrap())
+			}
+
+			resultChan <- result
+			// If there were no records on the pageRecord, we think that it was last pageRecord and exit.
+			if counter == 0 {
```
reugn, line 182: We can spare the last call by `counter < r.config.pageSize`.

io/aerospike/record_reader_paginated.go
```diff
@@ -0,0 +1,192 @@
...
+			r.logger.Debug("scan finished")
+			return nil, io.EOF
+		}
+
+		if res.result == nil {
+			return nil, io.EOF
+		}
+
+		if res.result.Err != nil {
+			r.logger.Error("error reading record", "error", res.result.Err)
```
reugn, line 63: Do we need to log here if we return the error?

io/aerospike/record_reader_paginated.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+// readPage reads the next record from pageRecord from the Aerospike database.
+func (r *RecordReader) readPage() (*models.Token, error) {
+	errChan := make(chan error)
+
+	if r.pageRecordsChan == nil {
+		r.pageRecordsChan = make(chan *pageRecord)
+		go r.startScanPaginated(errChan)
+	}
```
reugn, line 45: Looks a bit fragile to me. Why not initializing the scan in the constructor?

io/aerospike/record_reader.go
```diff
@@ -108,16 +116,27 @@ func NewRecordReader(
 	logger = logging.WithReader(logger, id, logging.ReaderTypeRecord)
 	logger.Debug("created new aerospike record reader")
 
-	return &RecordReader{
+	r := &RecordReader{
```
reugn, line 119: Can be reverted.

handler_backup_records.go
```diff
@@ -74,9 +75,14 @@ func (bh *backupRecordsHandler) run(
 		processors.NewVoidTimeSetter(bh.logger),
 		processors.NewTPSLimiter[*models.Token](
 			ctx, bh.config.RecordsPerSecond),
-	))
+	), bh.config.ParallelRead)
 
-	return pipeline.NewPipeline(readWorkers, composeProcessor, writers).Run(ctx)
+	pl, err := pipeline.NewPipeline(bh.config.SyncPipelines, readWorkers, composeProcessor, writers)
+	if err != nil {
+		return err
```
reugn, line 82: It's better to enrich the error with `failed to create a new pipeline: %w`

handler_backup.go
```diff
@@ -96,6 +99,20 @@ func newBackupHandler(
 	// redefine context cancel.
 	ctx, cancel := context.WithCancel(ctx)
 
+	var (
+		state *State
+		err   error
```
reugn, line 104: The error variable can be scoped under `if`.

handler_backup.go
```diff
@@ -96,6 +99,20 @@ func newBackupHandler(
...
+	)
+
+	if config.StateFile != "" {
+		// Keep in mind, that on continue operation, we update partitions list in config by pointer.
```
reugn, line 108: Could you elaborate on this. It is not clear from the comment.

cmd/internal/app/readers.go
```diff
@@ -26,11 +26,16 @@ import (
...
 	var opts []local.Opt
 
 	if c.Directory != "" && r.InputFile == "" {
-		opts = append(opts, local.WithDir(c.Directory), local.WithValidator(asb.NewValidator()))
+		opts = append(opts, local.WithDir(c.Directory))
+		// Append Validator only if backup params are not set.
+		// That means we don't need to check that we are saving a state file.
+		if b == nil {
+			opts = append(opts, local.WithValidator(asb.NewValidator()))
+		}
```
reugn, line 38: Consider taking this block to a method and reuse.

cmd/asbackup/readme.md
```diff
@@ -24,6 +24,9 @@ More info here https://goreleaser.com/quick-start/
 
 ## Supported flags
 ```
+Welcome to the Aerospike backup CLI tool!
+-----------------------------------------
+
```
reugn, line 29: Did we add this line to the usage output?

state.go
```diff
@@ -0,0 +1,271 @@
...
+	return ""
+}
+
+func openFile(ctx context.Context, reader StreamingReader, fileName string) (io.ReadCloser, error) {
+	readCh := make(chan io.ReadCloser)
+	errCh := make(chan error)
+
+	go reader.StreamFile(ctx, fileName, readCh, errCh)
+
+	for {
```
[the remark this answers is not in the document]
filkeith: removed

tests/integration/integration_test.go
filkeith: replaced through file

writers.go
filkeith: we already have `N` in `models.PartitionFilterSerialized`, this `n` is used to set number for worker. Later at [line 126](https://github.com/aerospike/backup-go/pull/158/files/5fe36eda95006d7a1c069e703dfa84cc8845a1c6#diff-883b695279ff275219de50e79984d2ac1068cd8d6c3a2fa004604994c8179325R126) we set it

state.go
filkeith: fixed

state.go
filkeith: removed, it was code from first implementation

state.go
filkeith: done

state.go
filkeith: changed

io/aerospike/record_reader_paginated.go
filkeith: added

io/aerospike/record_reader_paginated.go
filkeith: Don't think so, as we are using `scanPolicy.MaxRecords`, number of records is approximate and can be less

io/aerospike/record_reader_paginated.go
filkeith: fixed

io/aerospike/record_reader_paginated.go
filkeith: in constructor of `RecordReader` ? At the moment of creation of  `RecordReader` we don't know what reading will be used: normal or paginated. This is check for first run like [here](https://github.com/aerospike/backup-go/blob/main/io/aerospike/record_reader.go#L121)

io/aerospike/record_reader.go
filkeith: returned

handler_backup_records.go
filkeith: added

handler_backup.go
filkeith: moved

handler_backup.go
filkeith: moved setting filters code from state file to handler_backup

cmd/internal/app/readers.go
filkeith: you mean describe opts interface? and write some kind of setter for it? In each `new...Writer` func `opts` are different types

cmd/asbackup/readme.md
filkeith: it was there from the begining, I just wasn't coping full output to read me. Removed

cmd/internal/app/readers.go
reugn: Yes, this may introduce unnecessary complexity.

io/aerospike/record_reader_paginated.go
reugn: Yes, I see the complexity. Let's leave it as is for now.

writers.go
filkeith: fixed

reugn (approved):

PR #170, opened 2024-11-11
APPS-1329 Fix marker error for listing backups

filkeith:
- fixed marker parameter, renamed to startAfter
- added flags docs

### DIFF
--- a/io/azure/blob/options.go
+++ b/io/azure/blob/options.go
@@ -30,14 +30,10 @@ type options struct {
 	// When we stream files or delete files in folder, we skip directories. This flag will avoid skipping.
 	// Default: false
 	withNestedDir bool
-	// marker is a string value that identifies the portion of the list of containers to be
-	// returned with the next listing operation.
-	// The operation returns the NextMarker value within the response body if the listing
-	// operation did not return all containers remaining to be listed with the current page.
-	// The NextMarker value can be used
-	// as the value for the marker parameter in a subsequent call to request the next
-	// page of list items. The marker value is opaque to the client.
-	marker string
+	// startAfter is an artificial parameter. Used to skip objects in the storage.
+	// If it is set, then we compare the names received from the storage and if the name is less
+	// than the specified parameter, we skip this object.
+	startAfter string
 	// skipDirCheck if true, backup directory won't be checked.
 	skipDirCheck bool
 }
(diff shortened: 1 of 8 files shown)


filkeith:
After this merge, I'll update backup service

korotkov-aerospike (approved):

io/azure/blob/options.go
```diff
@@ -30,14 +30,10 @@ type options struct {
...
-	// marker is a string value that identifies the portion of the list of containers to be
-	// returned with the next listing operation.
-	// The operation returns the NextMarker value within the response body if the listing
-	// operation did not return all containers remaining to be listed with the current page.
-	// The NextMarker value can be used
-	// as the value for the marker parameter in a subsequent call to request the next
-	// page of list items. The marker value is opaque to the client.
-	marker string
+	// startAfter is an artificial parameter. Used to skip objects in the storage.
+	// If it is set, then we compare the names received from the storage and if the name is less
```
reugn, line 34: lexicographically

io/azure/blob/options.go
```diff
@@ -30,14 +30,10 @@ type options struct {
...
+	// than the specified parameter, we skip this object.
```
reugn, line 35: Mention that the boundary is exclusive.

io/azure/blob/reader.go
```diff
@@ -121,7 +120,8 @@ func (r *Reader) streamDirectory(
 
 		for _, blob := range page.Segment.BlobItems {
 			// Skip files in folders.
-			if isDirectory(r.prefix, *blob.Name) && !r.withNestedDir {
+			if (isDirectory(r.prefix, *blob.Name) && !r.withNestedDir) ||
+				isSkippedByStartAfter(r.startAfter, *blob.Name) {
```
reugn, line 124: It is better to extract the condition to a method and reuse.

io/azure/blob/options.go
filkeith: added

io/azure/blob/options.go
filkeith: added

io/azure/blob/reader.go
filkeith: fixed

reugn (approved):

PR #172, opened 2024-11-11
FMWK-597 Implement info commander for xdr requests

filkeith:
at the moment implemented only XDR commands as we don’t have MRT commands

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,8 @@
 module github.com/aerospike/backup-go
 
-go 1.21
+go 1.22
+
+toolchain go1.22.5
 
 require (
 	cloud.google.com/go/storage v1.43.0
@@ -35,6 +37,7 @@ require (
 	github.com/Azure/go-ansiterm v0.0.0-20210617225240-d185dfc1b5a1 // indirect
 	github.com/AzureAD/microsoft-authentication-library-for-go v1.2.2 // indirect
 	github.com/Microsoft/go-winio v0.6.2 // indirect
+	github.com/aerospike/aerospike-management-lib v1.5.0 // indirect
 	github.com/aws/aws-sdk-go-v2/aws/protocol/eventstream v1.6.4 // indirect
 	github.com/aws/aws-sdk-go-v2/credentials v1.17.28 // indirect
 	github.com/aws/aws-sdk-go-v2/feature/ec2/imds v1.16.12 // indirect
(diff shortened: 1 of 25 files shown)


io/xdr/info_commander.go
```diff
@@ -0,0 +1,209 @@
+package xdr
```
reugn, line 1: Consider moving the package under `io/aerospike`.

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/aerospike/backup-go
 
-go 1.21
+go 1.22
+
+toolchain go1.22.5
```
reugn, line 5: Revert the toolchain directive.

io/xdr/info_commander.go
```diff
@@ -0,0 +1,209 @@
...
+	cmdUnBlockMRTWrites = "namespaces"
+
+	cmdRespErrPrefix = "ERROR"
+)
+
+type InfoCommander struct {
+	client *info.AsInfo
+}
+
+func NewInfoCommander(host string, port int, user, password string) *InfoCommander {
```
reugn, line 31: Consider taking `host *aero.Host, policy *aero.ClientPolicy` as the parameters for more flexibility (e.g. future TLS support).

io/xdr/info_commander.go
```diff
@@ -0,0 +1,209 @@
...
+
+	if _, err = parseResultResponse(cmd, resp); err != nil {
+		return fmt.Errorf("failed to remove xdr dc: %w", err)
+	}
+
+	return nil
+}
+
+// BlockMRTWrites blocks MRT writes on cluster.
+func (c *InfoCommander) BlockMRTWrites(_, _ string) error {
```
reugn, line 167: Are those blank identifiers related to an interface signature?

io/xdr/info_commander.go
```diff
@@ -0,0 +1,209 @@
...
+
+	if _, err = parseResultResponse(cmd, resp); err != nil {
+		return fmt.Errorf("failed to block mrt writes: %w", err)
+	}
+
+	return nil
+}
+
+// UnBlockMRTWrites unblocks MRT writes on cluster.
+func (c *InfoCommander) UnBlockMRTWrites(_, _ string) error {
```
reugn, line 183: Are those blank identifiers related to an interface signature?

io/xdr/info_commander.go
```diff
@@ -0,0 +1,209 @@
...
+func (c *InfoCommander) createXDRDC(dc string) error {
+	cmd := fmt.Sprintf(cmdCreateXDRDC, dc)
+
+	resp, err := c.client.RequestInfo(cmd)
+	if err != nil {
+		return fmt.Errorf("failed to create xdr dc: %w", err)
+	}
+
+	if _, err = parseResultResponse(cmd, resp); err != nil {
+		return fmt.Errorf("failed to create xdr dc: %w", err)
```
reugn, line 85: nit: the majority of methods have the same error message string written twice

io/xdr/info_commander.go
filkeith: this blank identifiers are mocks for future MRT commands, at the moment I don't know how it will look like.  I've made them _ only because of linter.

go.mod
filkeith: fixed

io/xdr/info_commander.go
filkeith: fixed

PR #177, opened 2024-11-19
FMWK-610 Expose client policy timeouts

filkeith:
- added flags for client policy timeouts
- added separate flag for s3 bucket
- rearranged readme


cmd/internal/flags/common.go
```diff
@@ -19,48 +19,101 @@ import (
...
+	return &Common{operation: operation}
 }
 
 func (f *Common) NewFlagSet() *pflag.FlagSet {
 	flagSet := &pflag.FlagSet{}
 
+	var descNamespace, descSetList, descBinList, descNoRecords, descNoIndexes, descNoUDFs, descParallel string
+
+	switch f.operation {
+	case 0:
```
reugn, line 75: Could use the constants.

reugn (approved):

PR #183, opened 2024-11-25
FMWK-599 Rework pipelines for XDR support

filkeith:
Now we must configure routes for each pipeline step. There are 2 route rules that must cover all our needs.
- single mode - when only 1 channel is created for each stage. (It is how pipelines worked before)
- parallel mode - when we create a channel for each worker. (it is how pipelines work for continuation backup)

So now, if previous and next stage have same rule they are connected straight.
But if previous and next stage have different rules, we merg or split channels. For splitting channels, we must pass splitting function, that will return channel number (to which we must send a message).

P.S. I have to duplicate merge channel logic (made a copy from util), because without that Golang didn't allow me to use generic type from utils to use with generic type in pipeline.


pipeline/pipeline.go
```diff
@@ -40,62 +40,37 @@ type Worker[T any] interface {
...
+	stages []*stage[T]
 }
 
-var _ Worker[any] = (*Pipeline[any])(nil)
-
 const channelSize = 256
 
 // NewPipeline creates a new DataPipeline.
-func NewPipeline[T any](isSynced bool, workGroups ...[]Worker[T]) (*Pipeline[T], error) {
+func NewPipeline[T any](routes []Route[T], workGroups ...[]Worker[T]) (*Pipeline[T], error) {
```
reugn, line 49: Consider changing the constructor definition to:
```go
func NewPipeline[T any](sync bool, workGroups ...[]Worker[T]) (*Pipeline[T], error) {
	routes := newRoutes[T](sync, len(workGroups))
...
```
This way you encapsulate the work groups length calculation instead of requiring it from the caller.

pipeline/pipeline.go
```diff
@@ -40,62 +40,37 @@ type Worker[T any] interface {
...
 
-	// Check that all working groups have same number of workers.
-	if isSynced {
-		firstLen := len(workGroups[0])
-		for i := range workGroups {
-			if len(workGroups[i]) != firstLen {
-				return nil, fmt.Errorf("all workers groups must be same length in sync mode")
-			}
-		}
+	if len(routes) != len(workGroups) {
```
reugn, line 56: This check will also be redundant after the constructor refactoring.

pipeline/router.go
```diff
@@ -0,0 +1,252 @@
...
+		case prevOutMode == s.route.input.mode:
+			// If previous and next modes are the same, we connect workers directly.
+			output = r.create(s.route.output.mode, len(s.workers), s.route.output.bufferSize)
+		case prevOutMode == modeParallel && s.route.input.mode == modeSingle:
+			// Merge channels.
+			op := r.mergeChannels(prevOutput)
+			output = append(output, op)
+		case prevOutMode == modeSingle && s.route.input.mode == modeParallel:
+			// Split channels.
+			output = r.splitChannels(prevOutput[0], len(s.workers), s.route.output.sf)
```
reugn, line 133: Check that `s.route.output.sf` is not nil. Return an error otherwise.

pipeline/router.go
```diff
@@ -0,0 +1,252 @@
...
+			comChan := make(chan T, bufferSize)
+			result = append(result, comChan)
+		}
+	case modeSingle:
+		comChan := make(chan T, bufferSize)
+		result = append(result, comChan)
+	}
+
+	return result
+}
```
reugn, line 163: Consider refactoring to:
```go
func (r *router[T]) create(mode, workersNumber, bufferSize int) []chan T {
	size := workersNumber
	if mode == modeSingle {
		size = 1
	}
	result := make([]chan T, 0, size)
	for i := 0; i < size; i++ {
		comChan := make(chan T, bufferSize)
		result = append(result, comChan)
	}

	return result
}
```

pipeline/pipeline.go
```diff
@@ -182,25 +111,21 @@ func (dp *Pipeline[T]) Run(ctx context.Context) error {
 }
 
 type stage[T any] struct {
```
reugn, line 113: Document this internal type to improve code readability and understanding.

filkeith: done

pipeline/router.go
filkeith: done

pipeline/pipeline.go
filkeith: At the moment when we have only 2 types of route for pipeline it will work. But for MRT backup we will need different routes scheme like:
Screenshot 2024-11-27 at 12 26 27

So wee should have opportunity for flexible route configuration. So we will initialize MRT router in a different (third) way. That's why we can use one bool variable `sync`

pipeline/router.go
```diff
@@ -0,0 +1,252 @@
...
+		result = append(result, r)
+	}
+
+	return result
+}
+
+// RouteRule describes how exactly stages will communicate with each other.
+type RouteRule[T any] struct {
+	// mode can be single or parallel. Depending on that, one or more communication channels will be created.
+	mode int
```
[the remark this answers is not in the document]
filkeith: done

pipeline/router.go
```diff
@@ -0,0 +1,252 @@
...
+			// Many to one.
+			w.SetReceiveChan(input[j])
+			w.SetSendChan(output[0])
+		case len(input) < len(output) && len(input) == 1:
+			// One to many.
+			w.SetReceiveChan(input[0])
+			w.SetSendChan(output[j])
+		default:
+			return fmt.Errorf("failed to connect %d workers, with %d input and %d output",
+				len(st.workers), len(output), len(output))
```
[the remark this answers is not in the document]
filkeith: done

pipeline/router.go
```diff
@@ -39,6 +39,14 @@ type Route[T any] struct {
 	output *RouteRule[T]
 }
 
+func NewRoutes[T any](isSynced bool, stagesNum int) []Route[T] {
```
reugn, line 42: Is this constructor required? Please remove or add a doc line.

reugn (approved):

pipeline/router.go
filkeith: removed as unused

PR #185, opened 2024-11-26
FMWK-613 Restore user keys

filkeith:
- we skipped restore of PK on restore operation
- also we don't need to set sendKey on Scan:
`If the key is sent on a read, the server will generate the hash digest from the key and validate that digest with the digest sent by the client. Unless this is the explicit intent of the developer, avoid sending the key on reads. The default is to not send the user defined key.`


filkeith:
we have problem with default test server config, default user don't have correct rights

reugn (approved):

PR #186, opened 2024-11-27
FMWK-619 Implement lazy folder creation

filkeith:
- lazy folder creation
- copy send keys from write policy


io/aerospike/record_batch_writer.go
```diff
@@ -77,7 +77,7 @@ func (rw *batchRecordWriter) batchWrite(record *models.Record) *a.BatchWrite {
 
 func batchWritePolicy(writePolicy *a.WritePolicy, r *models.Record) *a.BatchWritePolicy {
 	policy := a.NewBatchWritePolicy()
-	policy.SendKey = true
+	policy.SendKey = writePolicy.SendKey
```
filkeith, line 80: Apply same value as in WritePolicy

PR #188, opened 2024-12-01
FMWK-620 Add directory list support for restore

filkeith:
regression tests passed


io/azure/blob/reader.go
```diff
@@ -253,3 +252,13 @@ func isSkippedByStartAfter(startAfter, fileName string) bool {
 
 	return false
 }
+
+// cleanPath is protection from incorrect input.
+func cleanPath(path string) string {
```
reugn, line 257: Can it be taken to utilities, or does it depend on the provider?

filkeith: for gcp and azure it will be the same, for aws it has additional check. Also didn't want to make dependencies to utils from io

PR #190, opened 2024-12-11
FMWK-598 Implement XDR data parser

filkeith:
- reworked info commands, moved them to existing asinfo package
- improved TCP server, to serve connections more efficiently
- XDR protocol parsed


io/aerospike/xdr/tcp_server.go
```diff
@@ -17,147 +17,401 @@ package xdr
...
+func (h *ConnectionHandler) handleMessages(ctx context.Context) {
+	parser := NewParser(h.conn)
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		default:
+			timeNow := atomic.LoadInt64(&h.timeNow)
+			deadline := timeNow + h.readTimoutNano
```
reugn, line 310: Those can be extracted to a method and reused.

reugn (approved):

PR #192, opened 2024-12-18
FMWK-629 Update documentation for the asrestore -r flag

filkeith:
- update documentation for -r flag on restore


filkeith:
@jyotigithub Please take a look if this descriptions is ok

jyotigithub:
Can we add following to explain -r behaviour along with -g or -u options?

This option still does a generation check by default and would need to be combined with the -g option if no generation check is desired. Note: this option is mutually exclusive to --unique.

filkeith:
@jyotigithub added, please take a look

reugn (approved):

PR #196, opened 2024-12-30
FMWK-603 Implement XDR backup handler

filkeith:
I'm still thinking about naming, so fill free to offer smth.
I've started to rework handlers, so in future (when will be adding backup to ASBX for ordinary Scan) we can easily refactor  current backup handlers:
- now file writers initialization is in separate `fileWriterProcessor`, so it can be reused for both handlers XDR and SCAN
- records reader initialization is in `recordReaderProcessor`, so it can be reused for both handlers XDR and SCAN
(for restore I'm planning to add `fileReaderProcessor`, and `recordWriterProcessor` )
- handler gather readers and writers and lunch pipeline


io_encoding.go
```diff
@@ -27,26 +28,29 @@ type EncoderType int
...
 // It is used to support different data formats.
 //
 //go:generate mockery --name Encoder
-type Encoder interface {
-	EncodeToken(*models.Token) ([]byte, error)
+type Encoder[T any] interface {
+	EncodeToken(T) ([]byte, error)
 	GetHeader() []byte
 	GenerateFilename(prefix, suffix string) string
 }
```
reugn, line 42: Consider adding a constraint type interface:
```go
type Token interface {
	models.Token | models.ASBXToken
}

// Encoder is an interface for encoding the types from the models package.
// It is used to support different data formats.
//
//go:generate mockery --name Encoder
type Encoder[T Token] interface {
	EncodeToken(*T) ([]byte, error)
	GetHeader() []byte
	GenerateFilename(prefix, suffix string) string
}
```
This will add type safety to the implementations.

config_backup_xdr.go
```diff
@@ -0,0 +1,79 @@
...
+	// DC name of dc that will be created on source instance.
+	DC string
+	// Local address, where source cluster will send data.
+	LocalAddress string
+	// Local port, where source cluster will send data.
+	LocalPort int
+	// Namespace is the Aerospike namespace to back up.
+	Namespace string
+	// Rewind is used to ship all existing records of a namespace.
+	// When rewinding a namespace, XDR will scan through the index and ship all the records for that namespace,
```
reugn, line 51: Format this long line.

filkeith:
need interim review

handler_backup_xdr.go
```diff
@@ -0,0 +1,191 @@
...
+// Wait waits for the backup job to complete and returns an error if the job failed.
+func (bh *HandlerBackupXDR) Wait(ctx context.Context) error {
+	defer func() {
+		bh.stats.Stop()
+	}()
+
+	select {
+	case <-bh.ctx.Done():
+		// Wait for global context.
+		return bh.ctx.Err()
```
reugn, line 166: If the global context was cancelled, this means that the handler completed successfully and nil should be returned (same issue in the regular handler).

reugn (commented):
The PR has only 17.16621% of patch coverage.

config_backup_xdr.go
```diff
@@ -0,0 +1,79 @@
...
+	// Max number of allowed simultaneous connection to server.
+	// Used by TCP server for XDR.
+	MaxConnections int
+	// How often a backup client will send info commands to check aerospike cluster stats.
+	// To measure recovery state and lag.
+	InfoPolingPeriodMilliseconds int64
+}
+
+func (cfg *ConfigBackupXDR) validate() error {
+	return nil
```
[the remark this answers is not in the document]
filkeith: done

handler_backup_xdr.go
```diff
@@ -0,0 +1,191 @@
...
+		return ctx.Err()
+	case err := <-bh.errors:
+		return err
+	}
+}
+
+// splitFunc distributes token between pipeline workers.
+func (bh *HandlerBackupXDR) splitFunc(t *models.ASBXToken) int {
+	partPerWorker := MaxPartitions / bh.config.ParallelWrite
+	id := t.Key.PartitionId() / partPerWorker
```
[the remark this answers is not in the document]
filkeith: done

internal/processors/token_counter.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"sync/atomic"
+)
+
+type TokenCounter[T any] struct {
+	counter *atomic.Uint64
+}
+
+// NewTokenCounter simple token counter.
```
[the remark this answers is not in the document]
filkeith: done

io_encoding.go
filkeith: done

config_backup_xdr.go
filkeith: done

handler_backup_xdr.go
filkeith: done

cmd/internal/app/asbackup.go
```diff
@@ -169,3 +228,11 @@ func (b *ASBackup) Run(ctx context.Context) error {
 
 	return nil
 }
+
+func getSecretAgent(b *backup.BackupConfig, bxdr *backup.ConfigBackupXDR) *backup.SecretAgentConfig {
+	if b != nil {
+		return b.SecretAgentConfig
+	}
+
+	return bxdr.SecretAgentConfig
```
reugn, line 237: Even though it is effectively not possible at the moment, it is worth checking `bxdr` for nil.

filkeith: fixed

reugn (approved):

PR #202, opened 2025-01-21
APPS-1471 Improve data validation

filkeith:
- Fix multiply dir validation
- Filter service sets

### DIFF
--- a/io/aerospike/xdr/tcp_server.go
+++ b/io/aerospike/xdr/tcp_server.go
@@ -112,13 +112,18 @@ func NewTCPServer(
 
 // Start launch tcp server for XDR.
 func (s *TCPServer) Start(ctx context.Context) (chan *models.ASBXToken, error) {
-	var err error
+	if !s.isActive.CompareAndSwap(false, true) {
+		return nil, errors.New("server start already initiated")
+	}
+
+	s.resultChan = make(chan *models.ASBXToken, s.config.ResultQueueSize)
 
 	// Redefine cancel function, so we can use it on Stop()
 	ctx, cancel := context.WithCancel(ctx)
 	s.cancel = cancel
 
 	// Create listener
+	var err error
 	if s.config.TLSConfig != nil {
 		s.listener, err = tls.Listen("tcp", s.config.Address, s.config.TLSConfig)
 		if err != nil {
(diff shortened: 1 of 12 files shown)


io/aerospike/xdr/tcp_server.go
```diff
@@ -153,9 +154,9 @@ func (s *TCPServer) Stop() {
 
 	s.wg.Wait()
 
-	if s.resultChan != nil {
+	s.once.Do(func() {
```
reugn, line 157: The current implementation does not allow starting the server after it has been stopped. Consider supporting this.

io/aerospike/xdr/tcp_server.go
```diff
@@ -379,7 +380,15 @@ func (h *ConnectionHandler) processMessage(ctx context.Context) {
 			}
 			// Create aerospike key.
 			key, err := NewAerospikeKey(aMsg.Fields)
-			if err != nil {
+
+			switch {
+			case err == nil:
+			// ok
```
reugn, line 386:
```suggestion
			     // ok
```

io/aerospike/xdr/tcp_server.go
```diff
@@ -379,7 +380,15 @@ func (h *ConnectionHandler) processMessage(ctx context.Context) {
...
+			case errors.Is(err, errSkipRecord):
+				// Make acknowledgement and skip record.
```
reugn, line 388:
```suggestion
				// Send acknowledgement and skip record.
```

io/aerospike/xdr/tcp_server.go
filkeith: replaced

io/aerospike/xdr/tcp_server.go
filkeith: fixed

io/aerospike/xdr/tcp_server.go
reugn: It is required to reset `resultChan` in `Start`.

io/aerospike/xdr/tcp_server.go
```diff
@@ -138,6 +139,8 @@ func (s *TCPServer) Start(ctx context.Context) (chan *models.ASBXToken, error) {
 		slog.String("address", s.config.Address),
 		slog.Bool("tls", s.config.TLSConfig != nil))
 
+	s.isActive.Store(true)
```
reugn, line 142: Check the current state in the `Start` and `Stop` methods and return an error if required.

io/aerospike/xdr/tcp_server.go
filkeith: moved chan initialization

reugn (approved):

PR #212, opened 2025-01-29
FMWK-651 Update documentation and improve test coverage

filkeith:
- reworked readme
- reworked readers

test coverage will be increased in separate task, when I'll finish adding new functionality.


io/aws/s3/reader.go
```diff
@@ -92,6 +84,11 @@ func NewReader(
 	r.client = client
 	r.bucketName = bucketName
 
+	// Presort files if needed.
+	if err := r.preSort(ctx); err != nil {
+		return nil, fmt.Errorf("failed to pre sort: %v", err)
```
reugn, line 89:
```suggestion
		return nil, fmt.Errorf("failed to pre sort: %w", err)
```

io/azure/blob/options.go
```diff
@@ -44,9 +37,8 @@ type options struct {
 	startAfter string
 	// skipDirCheck if true, backup directory won't be checked.
 	skipDirCheck bool
-	// Sort can be "", asc, desc (Use SortAsc, SortDesc constants).
-	// If sort is set, files will be sorted before read.
-	sort SortOrder
+	// isSorting shows if we need to sort files before read.
+	isSorting bool
```
reugn, line 41: Consider renaming to `requiresSorting` or `sortFiles`.

io/azure/blob/reader.go
```diff
@@ -75,17 +71,18 @@ func NewReader(
...
 	// Check if container exists.
 	if _, err := client.ServiceClient().NewContainerClient(containerName).GetProperties(ctx, nil); err != nil {
 		return nil, fmt.Errorf("unable to get container properties: %w", err)
 	}
 
 	r.containerName = containerName
 
+	// Presort files if needed.
+	if err := r.preSort(ctx); err != nil {
+		return nil, fmt.Errorf("failed to pre sort: %v", err)
```
reugn, line 83:
```suggestion
		return nil, fmt.Errorf("failed to pre sort: %w", err)
```

client.go
```diff
@@ -107,6 +107,7 @@ type Client struct {
 type ClientOpt func(*Client)
 
 // WithID sets the ID for the [Client].
+// Is used for loging.
```
reugn, line 110:
```suggestion
// This ID is used for logging purposes.
```

README.md
```diff
@@ -3,122 +3,513 @@
...
+
+- Standard backup and restore operations
+- MRT (Multi Record Transactions) through XDR (Cross-Datacenter Replication) backup support
+- Flexible backup configurations including:
+    - Partition-based backups
+    - Node-based backups
+    - Incremental backups using modification time filters
+    - Compression (ZSTD)
+    - Encryption (AES-128/256)
+    - Secret Agent integration
```
reugn, line 18:
```suggestion
    - [Secret Agent](https://aerospike.com/docs/tools/secret-agent) integration
```

io/azure/blob/reader.go
filkeith: fixed

client.go
filkeith: fixed

io/azure/blob/options.go
filkeith: fixed

io/aws/s3/reader.go
filkeith: fixed

reugn (approved):

PR #220, opened 2025-02-05
APPS-1509 Fix total records count

filkeith:
- skip counting MRT service set


internal/asinfo/info_client.go
```diff
@@ -42,6 +42,8 @@ const (
 	cmdUnBlockMRTWrites = "set-config:context=namespace;id=%s;disable-mrt-writes=false"
 
 	cmdRespErrPrefix = "ERROR"
+
+	setNameMRT = "<ERO~MRT"
```
reugn, line 46: Rename to `monitorRecordsSetName`.

internal/asinfo/info_client.go
```diff
@@ -761,6 +763,11 @@ func getRecordCountForNode(node infoGetter, policy *a.InfoPolicy, namespace stri
 			return 0, fmt.Errorf("set name missing in response %s", response[cmd])
 		}
 
+		// Skipping mrt service set.
```
reugn, line 766: > Skip MRT monitor records.

filkeith: fixed

reugn (approved):

PR #221, opened 2025-02-06
Fix getting backup files list for restore

filkeith:


io/aws/s3/reader.go
```diff
@@ -295,6 +295,12 @@ func (r *Reader) ListObjects(ctx context.Context, path string) ([]string, error)
 				continue
 			}
 
+			if r.validator != nil {
+				if err = r.validator.Run(*p.Key); err != nil {
```
reugn, line 299: `p.Key` can possibly be nil.

io/azure/blob/reader.go
```diff
@@ -270,6 +270,12 @@ func (r *Reader) ListObjects(ctx context.Context, path string) ([]string, error)
 				continue
 			}
 
+			if r.validator != nil {
+				if err = r.validator.Run(*blob.Name); err != nil {
```
reugn, line 274: `blob.Name` can possibly be nil.

reugn (approved):

PR #222, opened 2025-02-06
Fix parallelism for non bacth operations

filkeith:
- fixed parallelism for asbx and asb + disable-batch-writes
- fixed init of mixed restore
- fixed restore report (removed unused line)
- changed readers initializations for restore

### DIFF
--- a/cmd/internal/app/readers_test.go
+++ b/cmd/internal/app/readers_test.go
@@ -94,7 +94,7 @@ func createTmpFileLocal(dir, fileName string) error {
 	filePath := filepath.Join(dir, fileName)
 	f, err := os.Create(filePath)
 	if err != nil {
-		return fmt.Errorf("failed to create file: %v", err)
+		return fmt.Errorf("failed to create file: %w", err)
 	}
 	_ = f.Close()
 
--- a/cmd/internal/app/readers_test.go
+++ b/cmd/internal/app/readers_test.go
@@ -15,30 +15,48 @@
 package app
 
 import (
+	"bytes"
 	"context"
+	"fmt"
+	"os"
 	"path"
+	"path/filepath"
+	"strings"
 	"testing"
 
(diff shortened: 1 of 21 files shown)


cmd/internal/app/reports.go
```diff
@@ -73,15 +73,21 @@ func printRestoreReport(asbStats, asbxStats *bModels.RestoreStats) {
 	fmt.Println(headerRestoreReport)
 	fmt.Println(strings.Repeat("-", len(headerRestoreReport)))
 
+	// Swap vars if one of them null.
```
reugn, line 76: This seems like a hack. If there is no elegant way around this, explain why the swapping is required.

cmd/internal/app/readers_test.go
```diff
@@ -72,15 +90,29 @@ func TestNewLocalReader(t *testing.T) {
 	assert.Nil(t, reader)
 }
 
+func createTmpFileLocal(dir, fileName string) error {
+	filePath := filepath.Join(dir, fileName)
+	f, err := os.Create(filePath)
+	if err != nil {
+		return fmt.Errorf("failed to create file: %v", err)
```
reugn, line 97:
```suggestion
		return fmt.Errorf("failed to create file: %w", err)
```

io/common/storages.go
```diff
@@ -0,0 +1,5 @@
+package common
```
reugn, line 1: Consider giving the storage common package a better name, e.g. repository or just store. We can move more shared logic there.

filkeith: yep this package was created for future, when we will move logic there.

cmd/internal/app/readers_test.go
filkeith: fixed

reugn (approved):

PR #223, opened 2025-02-10
Add utility method to Sum backup and restore stats

korotkov-aerospike:


models/backup_stats.go
```diff
@@ -48,3 +50,16 @@ func (b *BackupStats) IsEmpty() bool {
 		b.GetSIndexes() == 0 &&
 		b.GetReadRecords() == 0
 }
+
+// SumBackupStats combines two BackupStats into a new one
+func SumBackupStats(a, b *BackupStats) *BackupStats {
```
filkeith, line 55: May be make it SumBackupStats(stats ...*BackupStats), and then sum all in for cycle

models/backup_stats.go
```diff
@@ -48,3 +50,16 @@ func (b *BackupStats) IsEmpty() bool {
...
+	result := &BackupStats{
+		commonStats: sumCommonStats(a.commonStats, b.commonStats),
```
filkeith, line 57: shold check a and b for nil first

filkeith (commented):
Should make stats sum more universal

filkeith (approved):
lgtm

PR #237, opened 2025-02-25
Add XDR max-throughput configuration flag

filkeith:


config_backup_xdr.go
```diff
@@ -171,6 +174,10 @@ func (c *ConfigBackupXDR) validate() error {
 		return fmt.Errorf("unsuported encoder type: %d", c.EncoderType)
 	}
 
+	if c.MaxThroughput%100 != 0 {
```
reugn, line 177: Add validation for non-negative values.

filkeith: added validation

reugn (approved):

PR #238, opened 2025-03-04
FMWK-688 Warm up connections before restore

filkeith:
- warm up connection. before restore

### DIFF
--- a/cmd/asrestore/readme.md
+++ b/cmd/asrestore/readme.md
@@ -123,6 +123,10 @@ Restore Flags:
                                   and each record is uploaded individually.
                                   The true max number of async Aerospike calls would then be <max-async-batches> * <batch-size>.
                                   Incompatible with --mode=asbx. (default 32)
+      --warm-up int               Warm Up fills the connection pool with connections for all nodes. This is necessary for batch restore.
+                                  At least must be same as (--max-async-batches + 1) or more, as one connection per node is reserved
+                                  for tend operations and is not used for transactions.
+                                  Incompatible with --mode=asbx. (default 33)
       --batch-size int            The max allowed number of records to simultaneously upload to Aerospike.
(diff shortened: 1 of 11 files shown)


cmd/internal/models/restore.go
```diff
@@ -27,12 +27,15 @@ type Restore struct {
...
 	BatchSize          int
 	MaxAsyncBatches    int
-	ExtraTTL           int64
-	IgnoreRecordError  bool
-	Uniq               bool
-	Replace            bool
-	NoGeneration       bool
-	TimeOut            int64
+	// Must be not less than MaxAsyncBatches.
+	// Will be applied only for batch writes.
```
reugn, line 31:
```suggestion
	// For optimal performance, should be at least MaxAsyncBatches.
	// This is applicable only to batch writes.
```

cmd/asrestore/readme.md
```diff
@@ -123,6 +123,10 @@ Restore Flags:
                                   and each record is uploaded individually.
                                   The true max number of async Aerospike calls would then be <max-async-batches> * <batch-size>.
                                   Incompatible with --mode=asbx. (default 32)
+      --warm-up int               Warm Up fills the connection pool with connections for all nodes. This is necessary for batch restore.
+                                  By default is calculated as (--max-async-batches + 1), as one connection per node is reserved
+                                  for tend operations and is not used for transactions.
+                                  Incompatible with --mode=asbx. (default 33)
```
reugn, line 129:
```suggestion
                                  Incompatible with --mode=asbx. (default 0)
```

filkeith: yep for 0 value it will be empty. So removed

cmd/internal/models/restore.go
filkeith: updated

reugn (approved):

PR #244, opened 2025-03-12
FMWK-690 Fix file size for compressed/encrypted files

filkeith:
- fix file size for compressed/encrypted files

### DIFF
--- a/io/counter/writer.go
+++ b/io/counter/writer.go
@@ -23,19 +23,27 @@ import (
 // We need it to get actual number of bytes written, after compression and encryption.
 type Writer struct {
 	writer io.WriteCloser
-	count  *atomic.Uint64
+	total  *atomic.Uint64
+	// The Counter is used for each file separately.
+	// It is reseted every time we create a new file.
+	fileSize *atomic.Uint64
 }
 
-func NewWriter(w io.WriteCloser, count *atomic.Uint64) *Writer {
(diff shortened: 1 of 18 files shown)


io/sized/writer.go
```diff
@@ -52,7 +55,7 @@ func NewWriter(ctx context.Context, n int, saveCommandChan chan int, limit int64
 }
 
 func (f *Writer) Write(p []byte) (n int, err error) {
-	if f.size >= f.limit {
+	if f.sizeCounter.Load() >= uint64(f.limit) {
```
reugn, line 58: It is better to have `f.limit` of `uint64` type in the first place.

io/counter/writer.go
```diff
@@ -23,19 +23,27 @@ import (
 // We need it to get actual number of bytes written, after compression and encryption.
 type Writer struct {
 	writer io.WriteCloser
-	count  *atomic.Uint64
+	total  *atomic.Uint64
+	// The Counter is used for each file separately.
+	// It is reseted every time we create a new file.
```
reugn, line 28:
```suggestion
	// This counter is used per backup file.
	// It is reset every time a new file is created.
```

io/sized/writer.go
filkeith: changed file limit to uint64

io/counter/writer.go
filkeith: updated

reugn (approved):

PR #245, opened 2025-03-13
FMWK-692 Fix PKCS1 key parsing

filkeith:
- fix parsing old style key

### DIFF
--- a/io_encryption.go
+++ b/io_encryption.go
@@ -25,11 +25,11 @@ import (
 
 const pemTemplate = "-----BEGIN PRIVATE KEY-----\n%s\n-----END PRIVATE KEY-----"
 
-// ReadPrivateKey parses and loads a private key according to the EncryptionPolicy
+// readPrivateKey parses and loads a private key according to the EncryptionPolicy
 // configuration. It can load the private key from a file, env variable or Secret Agent.
 // A valid agent parameter is required to load the key from Aerospike Secret Agent.
 // Pass in nil for any other option.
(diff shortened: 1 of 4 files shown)


io_encryption.go
```diff
@@ -78,6 +77,28 @@ func ReadPrivateKey(encPolicy *EncryptionPolicy, saConfig *SecretAgentConfig) ([
 	return sum256[:], nil
 }
 
+// parsePK parse private key from PKCS8 or PKCS1.
```
reugn, line 80:
```suggestion
// parsePK parses a private key from PKCS8 or PKCS1.
```

io_encryption.go
```diff
@@ -78,6 +77,28 @@ func ReadPrivateKey(encPolicy *EncryptionPolicy, saConfig *SecretAgentConfig) ([
...
+		return rsaKey, nil
+	}
+
+	// Try a PKCS1 format (which is always RSA).
+	pkcs1Key, err1 := x509.ParsePKCS1PrivateKey(block)
+	if err1 == nil {
+		return pkcs1Key, nil
+	}
+
+	return nil, fmt.Errorf("failed to parse RSA private key: %w (PKCS8), %w (PKCS1)", err8, err1)
```
reugn, line 99:
```suggestion
	return nil, errors.Join(err8, err1)
```
The error message is added in the caller function.

io_encryption.go
filkeith: fixed

io_encryption.go
filkeith: replaced

reugn (approved):

PR #256, opened 2025-03-27
FMWK-711 Implement metrics for pipeline

filkeith:


pipeline/pipeline.go
```diff
@@ -137,6 +138,14 @@ func (dp *Pipeline[T]) Run(ctx context.Context) error {
 	return nil
 }
 
+// GetMetrics returns stats: reader, writer.
+func (dp *Pipeline[T]) GetMetrics() (in, out int) {
+	_, readOut := dp.stages[0].GetMetrics()
+	writeIn, _ := dp.stages[2].GetMetrics()
```
reugn, line 144: Using fixed indexes is risky. The number of stages in the pipeline may vary.

pipeline/pipeline.go
```diff
@@ -30,6 +30,7 @@ type Worker[T any] interface {
 	SetSendChan(chan<- T)
 	SetReceiveChan(<-chan T)
 	Run(context.Context) error
+	GetMetrics() (int, int)
```
reugn, line 33: Why not returning `*models.Metrics` here. As the number of metrics grows, this will become inevitable.
You can add an `Add` method to `models.Metrics` for easier aggregation.

filkeith: Because In future we can add more different metrics from another entities. models.Metrics will be global struct with all available metrics. And current pipeline metric has only 2 ints, so creating a separate struct fro it will be overhead.

pipeline/pipeline.go
filkeith: fixed

models/metrics.go
```diff
@@ -0,0 +1,29 @@
...
+
+package models
+
+// Metrics contains app metrics.
+type Metrics struct {
+	PipelineReadQueueSize  int
+	PipelineWriteQueueSize int
+}
+
+// NewMetrics return metrics object with filled values.
```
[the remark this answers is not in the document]
filkeith: done

reugn (approved):

PR #263, opened 2025-04-14
FMWK-722 Add backup by rack support

filkeith:
- backup by rack lib support


cmd/internal/app/validation.go
```diff
@@ -210,6 +214,10 @@ func validateBackupParams(backupParams *models.Backup, commonParams *models.Comm
 		return fmt.Errorf("must specify either output-file or directory")
 	}
 
+	if backupParams.NodeList != "" && backupParams.RackList != "" {
```
korotkov-aerospike, line 217: isn't it "all" by default?

cmd/internal/app/validation.go
```diff
@@ -210,6 +214,10 @@ func validateBackupParams(backupParams *models.Backup, commonParams *models.Comm
...
+		return fmt.Errorf("must specify either rack-list or node-list")
```
korotkov-aerospike, line 218: "Specify either rack-list or node-list, but not both."
or something like this.
"either" implies that you need one one them.

handler_backup_records.go
```diff
@@ -35,6 +35,7 @@ import (
 type backupRecordsHandler struct {
 	config          *ConfigBackup
 	aerospikeClient AerospikeClient
+	infoClient      *asinfo.InfoClient
```
korotkov-aerospike, line 38: Maybe have interface here? To simplify mocks

korotkov-aerospike (approved):

PR #267, opened 2025-04-23
FMWK-730 Add kilobyte per second metric for reader and writer

filkeith:
- reworked metrics
- added kb per sec


internal/metrics/collector.go
```diff
@@ -0,0 +1,115 @@
...
+// Collector tracks and logs metrics such as request rate and counts within a context-managed environment.
+type Collector struct {
+	ctx     context.Context
+	enabled bool
+	name    string
+
+	Increment func()
+	Add       func(n uint64)
+
+	counter  atomic.Uint64
```
reugn, line 39: Consider renaming to `processed`, `total` or similar, since the metric now is not necessarily a counter.

internal/metrics/collector.go
```diff
@@ -0,0 +1,115 @@
...
+func (mc *Collector) report() {
+	ticker := time.NewTicker(time.Second)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case t := <-ticker.C:
+			count := mc.counter.Swap(0)
+			elapsed := t.Sub(mc.lastTime).Seconds()
+			rps := float64(count) / elapsed
```
reugn, line 85: Same here - not necessarily `rps`.

internal/metrics/collector.go
```diff
@@ -0,0 +1,115 @@
...
+			// for kbps metric we should divide by 1024.
+			if mc.name == MetricKilobytesPerSecond {
+				rps /= 1024
+			}
+
+			mc.lastResultMu.Lock()
+			mc.lastResult = rps
+			mc.lastResultMu.Unlock()
+
+			mc.logger.Debug("metrics, per second collector", slog.Float64(mc.name, rps))
```
reugn, line 96: It would be useful to have the operation type in the log (currently there is no unique id associated with each operation). Something like:
```suggestion
			mc.logger.Debug("backup metrics collector", slog.Float64(mc.name, rps))
```

config_backup.go
```diff
@@ -131,6 +131,9 @@ type ConfigBackup struct {
 	OutputFilePrefix string
 	// Retry policy for info commands.
 	InfoRetryPolicy *models.RetryPolicy
+
```
reugn, line 134:
```suggestion
```

cmd/internal/models/app.go
```diff
@@ -0,0 +1,21 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package models
+
+type App struct {
```
reugn, line 17: Add a doc for better understanding of the structure purpose.

handler_backup_xdr.go
```diff
@@ -51,7 +51,10 @@ type HandlerBackupXDR struct {
 
 	pl *pipeline.Pipeline[*models.ASBXToken]
 
-	rpsCollector *metrics.RPSCollector
+	// records per second collector.
+	rpsCollector *metrics.Collector
+	// bytes per second collector.
```
reugn, line 56: kilobytes

internal/metrics/collector_test.go
```diff
@@ -0,0 +1,264 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+const (
+	metricRecordsPerSecond   = "rps"
+	metricKilobytesPerSecond = "kbps"
+)
```
reugn, line 30: Use the exported constants from the package.

internal/metrics/collector.go
filkeith: fixed

internal/metrics/collector.go
filkeith: fixed

internal/metrics/collector.go
filkeith: This metrics is also used for restore, so I tried to make unified log string. And we can identify them by `mc.name`

config_backup.go
filkeith: fixed

cmd/internal/models/app.go
filkeith: Added comment to all the models

handler_backup_xdr.go

internal/metrics/collector_test.go
filkeith: fixed

internal/metrics/collector.go
reugn: You're right. `backup` was used as an example, it could have been `restore` as well. Without a unique identifier we may have mixed metric logs for several processes running simultaneously. Check if we can add a backup/restore marker, at least. If not, leave the message as `metrics collector` for now.

filkeith: I can add one more param to constructor, to identify

reugn (approved):

PR #270, opened 2025-04-28
FMWK-740 Improve metrics

filkeith:
- improved metrics

### DIFF
--- a/handler_restore.go
+++ b/handler_restore.go
@@ -238,6 +238,15 @@ func (rh *RestoreHandler[T]) Wait(ctx context.Context) error {
 
 // GetMetrics returns the rpsCollector of the backup job.
 func (rh *RestoreHandler[T]) GetMetrics() *models.Metrics {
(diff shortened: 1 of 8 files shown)


handler_restore.go
```diff
@@ -238,6 +238,15 @@ func (rh *RestoreHandler[T]) Wait(ctx context.Context) error {
 
 // GetMetrics returns the rpsCollector of the backup job.
 func (rh *RestoreHandler[T]) GetMetrics() *models.Metrics {
+	if rh.pl == nil {
```
korotkov-aerospike, line 241: I think, more robust solution is to init handler with not-nil metrics

filkeith: the problem is that pipeline doesn't exist, when you request metrics. So we must check if pl not nil, then return it metrics

korotkov-aerospike (dismissed):

reugn (approved):

PR #271, opened 2025-04-28
FMWK-741 Optimize backup records number estimation

korotkov-aerospike:
1) run estimate counting in goroutine
2) if scan by node — only scan one node
3) if scan by partition and filter by time/expression — only scan one partition
4) if full backup and scan by partition — use infocommand and calculate percentage.

### DIFF
--- a/config_backup.go
+++ b/config_backup.go
@@ -170,8 +170,7 @@ func (c *ConfigBackup) isStateContinue() bool {
 	return c.StateFile != "" && c.Continue
 }
 
-func (c *ConfigBackup) isFullBackup() bool {
-	// full backup doesn't have a lower bound.
+func (c *ConfigBackup) withoutFilter() bool {
 	return c.ModAfter == nil && c.ScanPolicy.FilterExpression == nil
 }
 
--- a/config_backup.go
+++ b/config_backup.go
@@ -172,7 +172,7 @@ func (c *ConfigBackup) isStateContinue() bool {
 
 func (c *ConfigBackup) isFullBackup() bool {
 	// full backup doesn't have a lower bound.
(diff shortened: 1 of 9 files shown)


config_backup.go
```diff
@@ -172,7 +172,7 @@ func (c *ConfigBackup) isStateContinue() bool {
 
 func (c *ConfigBackup) isFullBackup() bool {
```
filkeith, line 173: Now it is not full backup. Rename to smth like `isPartitionBackup`

handler_backup.go
```diff
@@ -319,6 +317,15 @@ func (bh *BackupHandler) backupSync(ctx context.Context) error {
 	return bh.recordHandler.run(ctx, writeWorkers, &bh.stats.ReadRecords)
 }
 
+func (bh *BackupHandler) countRecords(ctx context.Context) {
+	records, err := bh.recordHandler.countRecords(ctx, bh.infoClient)
+	if err != nil {
+		bh.logger.Error("failed to count records", "error", err)
```
filkeith, line 323: slog.Any("error", err)

filkeith (changes requested):

handler_backup_records_count.go
```diff
@@ -0,0 +1,154 @@
...
+		}
+
+		count++
+	}
+
+	return count * uint64(len(nodes)), nil
+}
+
+// randomPartition returns random partition from filters list.
+func randomPartition(partitionFilters []*a.PartitionFilter) *a.PartitionFilter {
```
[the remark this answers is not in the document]
korotkov-aerospike: I think every partition should have fair chance of selection

reugn (approved):

filkeith (approved):
lgtm

PR #280, opened 2025-05-07
FMWK-743 Fix RPS limiter for parallel mode

filkeith:
- fix rps limiter

### DIFF
--- a/handler_backup_records.go
+++ b/handler_backup_records.go
@@ -84,6 +84,9 @@ func (bh *backupRecordsHandler) run(
 		return err
 	}
 
+	// Calculate the Records Per Second (RPS) target for each individual parallel reader/processor.
+	// This value evenly distributes the overall read rate across the parallel workers,
+	// ensuring each worker adheres to a portion of the total RPS limit.
 	rps := bh.config.RecordsPerSecond / bh.config.ParallelRead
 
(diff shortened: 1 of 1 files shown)


handler_backup_records.go
```diff
@@ -84,11 +84,13 @@ func (bh *backupRecordsHandler) run(
 		return err
 	}
```
reugn, line 86: > // Calculate the Records Per Second (RPS) target for each individual parallel reader/processor. 
> // This value evenly distributes the overall read rate across the parallel workers, 
> // ensuring each worker adheres to a portion of the total RPS limit.

handler_backup_records.go
```diff
@@ -84,11 +84,13 @@ func (bh *backupRecordsHandler) run(
...
 
+	rps := bh.config.RecordsPerSecond / bh.config.ParallelRead
```
reugn, line 87: Is there validation to ensure `bh.config.ParallelRead` is positive?

filkeith: ParallelRead value is validated in backup config validation

handler_backup_records.go
filkeith: added

reugn (approved):

PR #291, opened 2025-05-22
FMWK-759 Fix rack-list backup for ASDB v6

filkeith:
- fix rack list backup for database v 6.4

### DIFF
--- a/internal/asinfo/info_client.go
+++ b/internal/asinfo/info_client.go
@@ -599,17 +599,17 @@ func (ic *InfoClient) GetRackNodes(rackID int) ([]string, error) {
 func (ic *InfoClient) getRackNodes(rackID int) ([]string, error) {
 	resp, err := ic.GetInfo(cmdRack)
 	if err != nil {
-		return nil, fmt.Errorf("failed get reacks info: %w", err)
+		return nil, fmt.Errorf("failed get racks info: %w", err)
 	}
 
 	result, err := parseResultResponse(cmdRack, resp)
 	if err != nil {
-		return nil, fmt.Errorf("failed to parse reacks info response: %w", err)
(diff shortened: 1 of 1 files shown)


internal/asinfo/info_client.go
```diff
@@ -604,12 +604,12 @@ func (ic *InfoClient) getRackNodes(rackID int) ([]string, error) {
 
 	result, err := parseResultResponse(cmdRack, resp)
 	if err != nil {
-		return nil, fmt.Errorf("failed to parse sets info response: %w", err)
+		return nil, fmt.Errorf("failed to parse reacks info response: %w", err)
```
reugn, line 607:
```suggestion
		return nil, fmt.Errorf("failed to parse racks info response: %w", err)
```

internal/asinfo/info_client.go
```diff
@@ -604,12 +604,12 @@ func (ic *InfoClient) getRackNodes(rackID int) ([]string, error) {
...
 	}
 
 	resultMap, err := parseInfoResponse(result, ";", ":", "=")
 	if err != nil {
-		return nil, fmt.Errorf("failed to parse sets info: %w", err)
+		return nil, fmt.Errorf("failed to parse reacks info: %w", err)
```
reugn, line 612:
```suggestion
		return nil, fmt.Errorf("failed to parse racks info: %w", err)
```

internal/asinfo/info_client.go
filkeith: fixed

internal/asinfo/info_client.go
filkeith: fixed

reugn (approved):

PR #295, opened 2025-06-04
FMWK-766 Skip empty files on restore

filkeith:
- skip empty files on restore

### DIFF
--- a/io/storage/azure/blob/reader.go
+++ b/io/storage/azure/blob/reader.go
@@ -188,7 +188,7 @@ func (r *Reader) streamDirectory(
 
 		for _, blobItem := range page.Segment.BlobItems {
 			if blobItem.Name == nil || blobItem.Properties == nil || blobItem.Properties.ContentLength == nil {
-				errorsCh <- fmt.Errorf("failed to get object attr: %s", path)
+				errorsCh <- fmt.Errorf("failed to get object attributes for %s", path)
 
 				return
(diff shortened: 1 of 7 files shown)


reugn (dismissed):

filkeith:
@reugn Added checks for Azure blob items

io/storage/azure/blob/reader.go
```diff
@@ -182,9 +182,16 @@ func (r *Reader) streamDirectory(
 		page, err := pager.NextPage(ctx)
 		if err != nil {
 			errorsCh <- fmt.Errorf("failed to get next page: %w", err)
+
+			return
 		}
 
 		for _, blobItem := range page.Segment.BlobItems {
+			if blobItem.Name == nil || blobItem.Properties == nil || blobItem.Properties.ContentLength == nil {
+				errorsCh <- fmt.Errorf("failed to get object attr: %s: %w", path, err)
```
reugn, line 191: The `err` object is not relevant here. I'm curious as to whether there is an option where the properties are not set, but the blob is still valid.

filkeith: I don't think that they can be nil, but added this check to avoid nil pointer error.
I think that all fields are pointers is just code style, that they use in lib. Also making BlobItem type unexported is very strange solution.

reugn (approved):

PR #303, opened 2025-06-18
FMWK-780 Udate go toolcahin to 1.24

filkeith:
- test


filkeith:
this should be done on new repo for cli

PR #305, opened 2025-06-25
APPS-1779 Fix limiter burst error

filkeith:
- added 8mb burst to limiter
- removed unused vars
- additional checks for readers on sending errors
- removed closing errors chan


io/storage/errors.go
```diff
@@ -14,10 +14,25 @@
...
+// If context is already canceled and the reader must be stopped, no need to send error to errors chan.
+func ErrToChan(ctx context.Context, ch chan<- error, err error) {
+	if err != nil {
+		select {
+		case <-ctx.Done():
+			return
+		case ch <- err:
+		}
+	}
+}
```
reugn, line 38: There is a possible race condition here, since select doesn't guarantee order. Consider:
```go
func ErrToChan(ctx context.Context, ch chan<- error, err error) {
	if err != nil && ctx.Err() == nil {
		select {
		case ch <- err:
		case <-ctx.Done():
		}
	}
}
```

shared.go
```diff
@@ -93,3 +93,11 @@ func newKeyByDigest(namespace, digest string) (*a.Key, error) {
 
 	return key, nil
 }
+
+func makeBandwidthLimiter(bandwidth int) *rate.Limiter {
```
reugn, line 97: nit: it is better to call it `newBandwidthLimiter` returning a pointer.

handler_backup_xdr.go
```diff
@@ -132,7 +132,8 @@ func newBackupXDRHandler(
 		infoClient:      infoClient,
 		stats:           stats,
 		logger:          logger,
-		errors:          make(chan error, 1),
+		errors:          make(chan error),
+		done:            make(chan struct{}),
```
reugn, line 136: Consider making buffered of size 1 to avoid possible blocking.

handler_backup_xdr.go
```diff
@@ -199,34 +200,36 @@ func (bh *HandlerBackupXDR) backup(ctx context.Context) error {
...
+	case <-bh.done: // Success
 	}
+
+	// Wait when all routines ended.
+	bh.wg.Wait()
+
+	// Clean.
+	bh.stats.Stop()
+	bh.rpsCollector.Stop()
+	bh.kbpsCollector.Stop()
```
reugn, line 230: Those fields are reused in different structs. Consider grouping and embedding. A single stats.Stop() can be exposed to encapsulate the inner logic.

shared.go
filkeith: agree

handler_backup_xdr.go

handler_backup_xdr.go
filkeith: moved

internal/processors/tps_limiter.go
```diff
@@ -29,7 +29,7 @@ type tpsLimiter[T models.TokenConstraint] struct {
 	tps     int
 }
 
-// NewTPSLimiter Create a new TPS limiter.
+// NewTPSLimiter Create a new TPS bandwidth.
```
reugn, line 32: Looks like an auto-rename issue.

internal/bandwidth/limiter.go
```diff
@@ -0,0 +1,56 @@
...
+package bandwidth
+
+import (
+	"context"
+
+	"golang.org/x/time/rate"
+)
+
+const (
+	// DefaultLimit is 8 mb as aerospike record size is limited with 8mb.
```
reugn, line 24: > DefaultLimit represents the minimum allowed bandwidth, constrained by the maximum record size (8 Mb).

internal/bandwidth/limiter.go
```diff
@@ -0,0 +1,56 @@
...
+	DefaultLimit = 8 * 1024 * 1024
+	// metaOverhead approximate size of record's metadata: namespace, set name, key, etc.
+	metaOverhead = 10_000
```
reugn, line 27:
```suggestion
	// metaOverhead represents an approximate size of record's metadata: namespace, set name, key, etc.
	metaOverhead = 10 * 1024
```

internal/bandwidth/limiter.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+// Limiter wrapper around standard go bandwidth.
```
reugn, line 30:
```suggestion
// Limiter wrapper around standard rate.Limiter.
```

internal/processors/tps_limiter.go
filkeith: fixed

reugn (approved):

PR #315, opened 2025-07-09
FMWK-788 Improve bandwidth limiter

filkeith:
New limiter implementation


internal/bandwidth/bucket.go
```diff
@@ -0,0 +1,100 @@
...
+
+	// After waiting, leak tokens again and consume.
+	rl.leak()
+	rl.tokens -= float64(n)
+}
+
+// leak updates the token count based on elapsed time.
+func (rl *Bucket) leak() {
+	now := time.Now()
+	elapsed := now.Sub(rl.lastLeak)
```
reugn, line 84:
```suggestion
	elapsed := time.Since(rl.lastLeak)
```

internal/bandwidth/bucket.go
```diff
@@ -0,0 +1,100 @@
...
+	rate float64
+
+	// Current available tokens (can be fractional for precise calculations).
+	tokens float64
+
+	// Last time we leaked tokens.
+	lastLeak time.Time
+}
+
+// NewBucket creates a new rate limiter with the specified limit and interval.
```
reugn, line 41: Will we benefit from a "burst" capability?

internal/bandwidth/limiter.go
```diff
@@ -15,53 +15,26 @@
...
-	bandwidth := MinLimit
-	if limit > bandwidth {
-		bandwidth = limit
-	}
-
-	bandwidth = int(float64(bandwidth)*base64Ratio) + metaOverhead
-
-	return bandwidth
+// Wait blocks until limiter permits n events to happen.
+func (l *Limiter) Wait(n int) {
```
reugn, line 38: It would be better to utilize the context.

internal/bandwidth/bucket.go
filkeith: I'm not sure if we need it at all

internal/bandwidth/bucket.go
```diff
@@ -0,0 +1,100 @@
...
+func NewBucket(limit int64, interval time.Duration) *Bucket {
```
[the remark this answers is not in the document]
filkeith: added check

internal/bandwidth/bucket.go
```diff
@@ -0,0 +1,100 @@
...
+		limit:    limit,
+		rate:     rate,
+		tokens:   float64(limit),
+		lastLeak: time.Now(),
+	}
+}
+
+// Wait blocks until n tokens are available.
+// It allows waiting for amounts larger than the limit.
+func (rl *Bucket) Wait(n int64) {
```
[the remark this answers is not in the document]
filkeith: added

internal/bandwidth/bucket.go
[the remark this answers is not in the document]
filkeith: but I have overflow check below
```
		if rl.tokens > float64(rl.limit) {
			rl.tokens = float64(rl.limit)
		}
```

reugn (approved):

PR #322, opened 2025-08-03
APPS-1820 Rework file writer closing mechanism

filkeith:
**How it was:**
Because we write indexes and udfs to the first file, and we need first file to be opened to write records data to it, old implementation was confusing. We had io.Writer interface, with empty close() method, so pipeline didn't close anything at the end of processing, it calls Clsoe() that does nothing. And then we manually closed all writers in backup handler.

**Now:**
May be it's not most elegant solution, so I'm opened for proposals. 
I made wrapper for writer, with no-op Close(). So now, when we backup indexes and udfs we use writer wrapped with `noCloseWriter`. For all other operations we use normal io.WriterCloser, so when pipeline finish processing file it realy close it.


handler_backup.go
```diff
@@ -397,18 +395,12 @@ func (bh *BackupHandler) countRecords(ctx context.Context) {
...
-			logger.Error("failed to close backup file", "error", err)
-		}
-	}
-}
-
 func (bh *BackupHandler) backupSIndexesAndUDFs(
 	ctx context.Context,
 	writer io.WriteCloser,
 ) error {
+	writer = newNoCloseWriter(writer)
```
reugn, line 402: > // The original writer is wrapped to disable closing after writing metadata.

token_writers.go
```diff
@@ -35,6 +35,7 @@ type statsSetterToken interface {
 	AddSIndexes(uint32)
 }
 
+// tokenStatsWriter used to count UDFs and SIndexes.
```
reugn, line 38:
```suggestion
// tokenStatsWriter is used to count UDFs and SIndexes.
```

token_writers_test.go
```diff
@@ -29,6 +30,18 @@ import (
...
 
+type bufferWriteCloser struct {
+	*bytes.Buffer
+}
+
+func (bwc *bufferWriteCloser) Close() error {
+	return nil // no-op
+}
+
+func NewBufferWriteCloser(buf *bytes.Buffer) io.WriteCloser {
```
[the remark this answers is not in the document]
filkeith: fixed

token_writers.go
```diff
@@ -157,9 +157,37 @@ func (w *tokenWriter[T]) Write(ctx context.Context, v T) (int, error) {
...
+}
+
+// newNoCloseWriter wraps an io.Writer, disabling the Close operation while preserving Write functionality.
+func newNoCloseWriter(w io.Writer) io.WriteCloser {
+	return &noCloseWriter{
+		writer: w,
+	}
+}
+
+// Write writes bytes to writer.
```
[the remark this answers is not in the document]
filkeith: fixed

processor_file_writer.go
```diff
@@ -87,27 +87,41 @@ func newFileWriterProcessor[T models.TokenConstraint](
...
 	return dataWriters
 }
 
+// createDataWriter creates a single data writer with state info if available.
+func (fw *fileWriterProcessor[T]) createDataWriter(writer io.WriteCloser, n int) pipe.Writer[T] {
+	sInfo := fw.getStateInfo(n)
+	tWriter := newTokenWriter(
+		fw.encoder,
+		writer,
+		fw.logger.With(slog.String("writer", fmt.Sprintf("%d", n))),
```
reugn, line 109:
```suggestion
		fw.logger.With(slog.Int("writer", n)),
```

reugn (approved):

PR #326, opened 2025-08-12
FMWK-804 Improve partition filter validation

filkeith:
- added one more validation
- updated errors text

### DIFF
--- a/config_partition_filter.go
+++ b/config_partition_filter.go
@@ -79,7 +79,11 @@ func NewPartitionFilterAll() *a.PartitionFilter {
 
 // splitPartitions splits the partitions to groups.
 func splitPartitions(partitionFilters []*a.PartitionFilter, numWorkers int) ([]*a.PartitionFilter, error) {
-	if numWorkers < 1 || numWorkers < len(partitionFilters) {
+	if numWorkers < 1 {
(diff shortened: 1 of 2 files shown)


config_partition_filter.go
```diff
@@ -165,7 +170,11 @@ func splitPartitions(partitionFilters []*a.PartitionFilter, numWorkers int) ([]*
 }
 
 // splitPartitionRange splits one range filter to numWorkers.
-func splitPartitionRange(partitionFilters *a.PartitionFilter, numWorkers int) []*a.PartitionFilter {
+func splitPartitionRange(partitionFilters *a.PartitionFilter, numWorkers int) ([]*a.PartitionFilter, error) {
+	if partitionFilters.Count < numWorkers {
```
reugn, line 174: The condition doesn't correspond to the error message.

config_partition_filter.go
```diff
@@ -80,28 +80,28 @@ func NewPartitionFilterAll() *a.PartitionFilter {
 // splitPartitions splits the partitions to groups.
 func splitPartitions(partitionFilters []*a.PartitionFilter, numWorkers int) ([]*a.PartitionFilter, error) {
 	if numWorkers < 1 || numWorkers < len(partitionFilters) {
```
reugn, line 82: It is better to have a separate error message for `numWorkers < 1`.

PR #335, opened 2025-09-04
FMWK-797 Add support for expression-based indexes backup

filkeith:


filkeith:
updated from main

io/encoding/asb/decode.go
```diff
@@ -406,15 +434,32 @@ func (r *Decoder[T]) readGlobals() (any, error) {
...
+			if err = r.skipToNextLine(); err != nil {
+				return nil, fmt.Errorf("failed to skip unknown global line type %c: %w", b, err)
+			}
+
+			r.logger.Warn("ignoring error while reading global type",
+				slog.Any("error", fmt.Errorf("invalid global line type %c", b)))
+
+			// Recursively try to read next global
+			return r.readGlobals()
+		}
```
reugn, line 448: Consider extracting to a method and reusing. Also, it would be better to limit the recursion calls on an error.

pkg/asinfo/client.go
```diff
@@ -190,6 +190,23 @@ func (ic *Client) GetVersion() (AerospikeVersion, error) {
...
+	list, err := ic.GetSIndexes(namespace)
+	if err != nil {
+		return false, err
+	}
+
+	for i := range list {
+		if list[i].Expression != "" {
+			return true, nil
+		}
+	}
```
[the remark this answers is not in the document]
filkeith: fixed

io/encoding/asb/version_test.go
```diff
@@ -0,0 +1,326 @@
...
+// limitations under the License.
+
+package asb
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestParseVersion(t *testing.T) {
```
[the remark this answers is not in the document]
filkeith: fixed

io/encoding/asb/decode.go
filkeith: fixed, also adding limit to recursion will lead to stop processing on future errors. We skip line and only then recursively read. Also it's not really recursion, we just run function once again on new line

handler_backup.go
```diff
@@ -64,6 +64,7 @@ type BackupHandler struct {
 	readerProcessor *recordReaderProcessor[*models.Token]
 	writerProcessor *fileWriterProcessor[*models.Token]
 	encoder         Encoder[*models.Token]
+	metaEncoder     Encoder[*models.Token]
```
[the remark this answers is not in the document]
filkeith: Inside pipeline we don't know what we are writing at the moment: records or meta data, so don't know what header to use: with version 3.1 or 3.1. Because of that I init separate encoder and run separate pipeline for metadata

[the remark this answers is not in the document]
reugn: Just use the latest for both and skip unknown fields if configured to do so. Need to check the performance implications of introducing support for v3.2.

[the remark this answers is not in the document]
filkeith: removed

io/encoding/asb/decode_test.go
```diff
@@ -75,7 +76,7 @@ func TestASBReader_readHeader(t *testing.T) {
 			wantErr: true,
 		},
 		{
-			name: "negative bad version number",
+			name: "negative bad metaVersion number",
```
reugn, line 79: It's not a meta version, but rather a file/format.

io/encoding/asb/version.go
```diff
@@ -0,0 +1,92 @@
...
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+var (
+	// versionDefault is the default version of the ASB format.
+	versionDefault = newVersion(3, 1)
```
reugn, line 25: Consider renaming the version variables to `version31, version32..` for better readability.

io/encoding/asb/encode.go
```diff
@@ -107,13 +110,18 @@ func (e *Encoder[T]) encodeSIndex(sindex *models.SIndex, buff *bytes.Buffer) (in
 	return sindexToASB(sindex, buff)
 }
 
-func (e *Encoder[T]) GetHeader(_ uint64) []byte {
+func (e *Encoder[T]) GetHeader(_ uint64, isMeta bool) []byte {
```
reugn, line 113: It is better to pass as a parameter the exact version string.

io/encoding/asb/decode.go
```diff
@@ -280,7 +290,7 @@ func (r *Decoder[T]) readHeader() (*header, error) {
 		return nil, err
 	}
 
-	// version number format is "x.y"
+	// metaVersion number format is "x.y"
```
reugn, line 293: metaVersion -> asb version

io/encoding/asb/decode.go
```diff
@@ -207,8 +212,13 @@ func NewDecoder[T models.TokenConstraint](src io.Reader, fileName string) (*Deco
 		return nil, fmt.Errorf("error while reading %s header: %w", fileName, err)
 	}
 
-	if asb.header.Version != supportedVersion {
-		return nil, fmt.Errorf("unsupported backup file version: %s", asb.header.Version)
+	fileVersion, err := parseVersion(asb.header.Version)
+	if err != nil {
+		return nil, fmt.Errorf("error while parsing %s header metaVersion: %w", fileName, err)
```
reugn, line 217: asb header version?

handler_backup_xdr.go
```diff
@@ -65,15 +65,21 @@ func newBackupXDRHandler(
...
+) (*HandlerBackupXDR, error) {
 	id := uuid.NewString()
 	logger = logging.WithHandler(logger, id, logging.HandlerTypeBackup, writer.GetType())
 	metricMessage := fmt.Sprintf("%s metrics %s", logging.HandlerTypeBackup, id)
 
 	// redefine context cancel.
 	ctx, cancel := context.WithCancel(ctx)
 
-	encoder := NewEncoder[*models.ASBXToken](config.EncoderType, config.Namespace, false)
+	hasExprSind, err := infoClient.HasExpressionSindex(config.Namespace)
```
reugn, line 76: `hasExpSIndex` for better readability.

handler_backup.go
```diff
@@ -137,7 +137,13 @@ func newBackupHandler(
 		}
 	}
 
-	encoder := NewEncoder[*models.Token](config.EncoderType, config.Namespace, config.Compact)
+	hasExprSind, err := infoClient.HasExpressionSindex(config.Namespace)
+	if err != nil {
+		cancel()
```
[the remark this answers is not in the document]
filkeith: I'm using this cancel in Wait() function, so I need it in constructor

io/encoding/asb/decode_test.go
filkeith: reverted comment

io/encoding/asb/encode.go
filkeith: then I need to move version define logic from encoder somewhere to backup handler, I don't want to do that because we have different encoders, and they should encapsulate their logic

handler_backup_xdr.go

io/encoding/asb/version.go
filkeith: fixed

io/encoding/asb/decode_test.go
```diff
@@ -1725,16 +1726,85 @@ func TestASBReader_readRecord(t *testing.T) {
...
+			name: "unknown field",
+			fields: fields{
+				reader: newTestCountingReader(
+					"+ k I 10\n" +
+						"+ n namespace1\n" +
+						"+ d " + encodedDigest + "\n" +
+						"+ s set1\n" +
+						"unknown\n" +
+						"+ g 10\n" +
+						"+ t 999999999999999999999999999999999999\n" +
```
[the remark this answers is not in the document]
filkeith: but it is already between field: `"unknown\n" +`
t 99999... is valid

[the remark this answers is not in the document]
filkeith: fixed this test

PR #339, opened 2025-10-03
Refine backup size estimation

reugn:
Account for non-default `FileLimit` configuration when calculating headers size.


filkeith (approved):
lgtm

PR #371, opened 2025-12-25
FMWK-882 Fix goroutine leak on successful backup completion

reugn:
Ensures `BackupHandler.Wait()` cancels the context when backup succeeds, allowing State goroutines to terminate and preventing resource leaks.

### DIFF
--- a/handler_backup.go
+++ b/handler_backup.go
@@ -21,7 +21,6 @@ import (
 	"io"
 	"log/slog"
 	"math"
-	"sync"
 	"sync/atomic"
 
 	"github.com/aerospike/backup-go/internal/bandwidth"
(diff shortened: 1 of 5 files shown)


korotkov-aerospike:
I think, the root problem is that we store context in handler.

func (bh *BackupHandler) run() {
	bh.wg.Add(1)
	bh.stats.Start()

	go doWork(bh.errors, bh.done, bh.logger, func() error {
		defer bh.wg.Done()

		return bh.backup(bh.ctx)
	})
}



Why not change signature to 

func (bh *BackupHandler) run(ctx context.context) {

?

filkeith:
@korotkov-aerospike 

>I think, the root problem is that we store context in handler.

We can change
`run()` to `run(ctx context,Context)` 
but we need to save context to handler anyway to check on `Wait()`.

filkeith (approved):
lgtm

PR #428, opened 2026-04-01
BKRS-155 Optimize memory management for S3 backup

korotkov-aerospike:
Test shows that new implementation works as fast as an old one, reducing memory usage 11-40%.

### DIFF
--- a/internal/util/collections/byte_buffer_pool.go
+++ b/internal/util/collections/byte_buffer_pool.go
@@ -39,14 +39,15 @@ func NewByteBufferPool(size int) *ByteBufferPool {
 // Get retrieves a buffer from the pool or allocates one.
 // The buffer is always reset to an empty state.
 func (p *ByteBufferPool) Get() *bytes.Buffer {
-	slog.Warn("Getting byte buffer pool")
 	b := p.pool.Get().(*bytes.Buffer)
 	b.Reset()
 
+	slog.Warn("Getting byte buffer pool", slog.Int("size", b.Cap()))
 	return b
 }
 
 // Put returns a buffer to the pool.
 func (p *ByteBufferPool) Put(b *bytes.Buffer) {
+	slog.Warn("Putting byte buffer pool", slog.Int("size", b.Cap()))
 	p.pool.Put(b)
 }
--- a/internal/util/collections/byte_buffer_pool.go
+++ b/internal/util/collections/byte_buffer_pool.go
@@ -16,6 +16,7 @@ package collections
 
 import (
 	"bytes"
+	"log/slog"
 	"sync"
 )
 
@@ -28,6 +29,7 @@ type ByteBufferPool struct {
 func NewByteBufferPool(size int) *ByteBufferPool {
 	p := &ByteBufferPool{}
 	p.pool.New = func() any {
+		slog.Warn("Creating new byte buffer pool")
 		return bytes.NewBuffer(make([]byte, 0, size))
 	}
 
@@ -37,6 +39,7 @@ func NewByteBufferPool(size int) *ByteBufferPool {
 // Get retrieves a buffer from the pool or allocates one.
 // The buffer is always reset to an empty state.
 func (p *ByteBufferPool) Get() *bytes.Buffer {
+	slog.Warn("Getting byte buffer pool")
 	b := p.pool.Get().(*bytes.Buffer)
 	b.Reset()
 
--- a/internal/util/collections/byte_buffer_pool.go
+++ b/internal/util/collections/byte_buffer_pool.go
@@ -16,7 +16,6 @@ package collections
 
 import (
 	"bytes"
-	"log/slog"
 	"sync"
 )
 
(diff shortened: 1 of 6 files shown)


internal/util/collections/byte_buffer_pool.go
```diff
@@ -0,0 +1,49 @@
...
+func (p *ByteBufferPool) Get() *bytes.Buffer {
+	b := p.pool.Get().(*bytes.Buffer)
+	b.Reset()
+
+	return b
+}
+
+// Put returns a buffer to the pool.
+func (p *ByteBufferPool) Put(b *bytes.Buffer) {
+	p.pool.Put(b)
```
filkeith, line 48: if our there will be more data than size, buffer's cap can grow, and on next cycle it will have cap * 2. So we should abandon such buffers, and let gc to collect them, so add check:

if b.Cap() > size {
  return
}

io/storage/aws/s3/writer.go
```diff
@@ -210,7 +215,12 @@ type s3Writer struct {
 
 var _ io.WriteCloser = (*s3Writer)(nil)
 
+var bufferPool = collections.NewByteBufferPool(s3DefaultChunkSize)
```
filkeith: I'm not sure that we should make it global, it will do the same job inside the writer. As each writer uses more than 1 buffer at one moment on time.

filkeith (changes requested):

io/storage/aws/s3/writer.go
```diff
@@ -219,19 +237,64 @@ func (w *s3Writer) Write(p []byte) (int, error) {
...
+		// This ensures that after this call, Cap >= chunkSize
+		slog.Info("Grow buffer", slog.Int("to", w.chunkSize-w.buffer.Len()))
+		w.buffer.Grow(w.chunkSize - w.buffer.Len())
 	}
 
-	return w.buffer.Write(p)
+	// 3. The Write Loop
+	written := 0
+
+	for len(p) > 0 {
```
filkeith, line 264: What if we avoid buffer at all. We already copy value to uploadPart() and it contains a copy inside func with pointer on stack. May be we can move all async logic inside uploadPart(). So no buffers will be nedded

korotkov-aerospike:
* check if we can just use uplaod mamanger.

io/storage/aws/s3/writer.go
korotkov-aerospike: We don't copy there, we pass pointer to bytes in buffer.

io/storage/aws/s3/writer.go
```diff
@@ -200,38 +214,99 @@ type s3Writer struct {
...
+	w.writeMu.Lock()
+	defer w.writeMu.Unlock()
+
 	if w.closed.Load() {
 		return 0, os.ErrClosed
 	}
 
-	if err := w.uploadErr.Load(); err != nil {
-		return 0, err.(error)
+	if err := w.firstUploadErr(); err != nil {
```
filkeith: Why you need this method? you can call here w.uploadErr.Load() and return pointer

io/storage/aws/s3/writer.go
```diff
@@ -200,38 +214,99 @@ type s3Writer struct {
...
-		})
-		// Reset buffer for the next chunk.
-		w.buffer = new(bytes.Buffer)
+	if w.buffer == nil {
+		if w.bufferPool != nil {
+			w.buffer = w.bufferPool.Get()
+		} else {
+			w.buffer = bytes.NewBuffer(make([]byte, 0, w.chunkSize))
+		}
 	}
```
filkeith: Mey be move it to separate method

func (w *Writer) newBuffer() *bytes.Buffer {
  if w.bufferPool != nil {
    return w.bufferPool.Get()
  }
   
  return bytes.NewBuffer(make([]byte, 0, w.chunkSize))
}

so it will be called like

if w.buffer == nil {
   w.buffer = w.newBuffer()
}

it will make it easy to read

io/storage/aws/s3/writer.go
```diff
@@ -200,38 +214,99 @@ type s3Writer struct {
...
 
-	return w.buffer.Write(p)
+	// Buffer Promotion. If we have a small buffer, grow it.
+	if w.buffer.Cap() < w.chunkSize {
+		// This ensures that after this call, Cap >= chunkSize
+		w.buffer.Grow(w.chunkSize - w.buffer.Len())
+	}
+
+	// 3. The Write Loop
```
filkeith: remove this number

io/storage/aws/s3/writer.go
```diff
@@ -288,42 +364,40 @@ func (w *s3Writer) Close() error {
...
+		}
+		w.buffer = nil
 	}
+
 	// Wait for all workers to finish.
 	w.workersPool.Wait()
 
-	if err := w.uploadErr.Load(); err != nil {
-		return err.(error)
+	if err := w.firstUploadErr(); err != nil {
```
filkeith: same here, it is function with 1 line, why wrapping it?

io/storage/aws/s3/writer.go
korotkov-aerospike: Need.
uploadErr   atomic.Pointer[error] holds pointer to an error.

func (w *s3Writer) getUploadErr() error {
	if p := w.uploadErr.Load(); p != nil {
		return *p <------ dereference
	}

	return nil <------- no dereference
}

filkeith (approved):
lgtm

PR #431, opened 2026-04-12
Encode token memory optimization

korotkov-aerospike:
Impact: Slight memory improvement.

Analysis: This approach consistently shaved about 30–40 MB off the peak memory compared to the Baseline (dropping from ~1.08 GB to ~1.04 GB).

Cursor:
  1. API change `EncodeToken(T) ([]byte, error)` → `EncodeToken(T, io.Writer) error` - Good. Pushing the writer down lets callers control allocation. The
     caller decides whether to use a pooled buffer, pre-allocated buffer, or write directly.
  2. Removing `estimatedRecordSize` heuristic in ASB encoder - Fine. The smoothed size estimation was a micro-optimization that added complexity. If you're
     reusing buffers at the caller level, this doesn't help much anyway.
  3. Removing wrapper methods `encodeRecord`, `encodeUDF`, `encodeSIndex` - Good cleanup. They were just pass-throughs adding indirection.

### DIFF
--- a/io/encoding/asbx/encode.go
+++ b/io/encoding/asbx/encode.go
@@ -15,8 +15,8 @@
 package asbx
 
 import (
+	"bytes"
 	"fmt"
-	"io"
 	"sync/atomic"
 
 	"github.com/aerospike/backup-go/models"
@@ -53,7 +53,7 @@ func (e *Encoder[T]) GenerateFilename(prefix, suffix string) string {
 }
 
 // EncodeToken encodes a token to the ASBX format, writing to the provided writer.
-func (e *Encoder[T]) EncodeToken(token T, w io.Writer) error {
+func (e *Encoder[T]) EncodeToken(token T, w *bytes.Buffer) error {
 	t, ok := any(token).(*models.ASBXToken)
 	if !ok {
 		return fmt.Errorf("unsupported token type %T for ASBX encoder", token)
--- a/io/encoding/asbx/encode.go
+++ b/io/encoding/asbx/encode.go
@@ -16,6 +16,7 @@ package asbx
 
 import (
 	"fmt"
+	"io"
 	"sync/atomic"
 
 	"github.com/aerospike/backup-go/models"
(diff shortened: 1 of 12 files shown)


io/encoding/asbx/encode.go
```diff
@@ -51,36 +52,43 @@ func (e *Encoder[T]) GenerateFilename(prefix, suffix string) string {
...
 	}
 	// Message contains:
 	// Digest - 20 bytes.
 	// Payload Size - 6 bytes.
 	// Payload - contains a raw message from tcp protocol.
 	pLen := len(t.Payload)
 
-	msg := make([]byte, 26+pLen)
+	// Write digest (20 bytes).
+	if _, err := w.Write(t.Key.Digest()); err != nil {
```
filkeith, line 68: Move this write to others, also ma be we can aggregate them in to one write... Like right now we can get partial record write if error happens in the middle

token_writers.go
```diff
@@ -129,8 +131,9 @@ func newTokenWriter[T models.TokenConstraint](
 
 // Write encodes v and writes it to the output.
 func (w *tokenWriter[T]) Write(v T) (int, error) {
-	data, err := w.encoder.EncodeToken(v)
-	if err != nil {
+	w.buf.Reset()
```
filkeith, line 134: I missed smth ,ay be. But the whole PR is about removing unnecessary allocations and writing directly to the writer. But here you add buffer.

io/encoding/asbx/encode.go
korotkov-aerospike: w.buf.Reset()

	if err := w.encoder.EncodeToken(v, &w.buf); err != nil {
		return 0, fmt.Errorf("failed to encode token: %w", err)
	}

Another reason why we should write to a buffer. If write fails - it's only 	w.buf corrupted (single record)

filkeith (approved):
Looks good. Hope all tests passes, also data integrity tests, as we can't allow data corruption

PR #432, opened 2026-04-12
Encryption memory optimization

korotkov-aerospike:
WITH optimization

Backup report
-------------
Start Time:           Mon, 27 Apr 2026 13:04:06 UTC
Duration:             1m15.696456572s

Records Read:         30000000
sIndex Read:          0
UDFs Read:            0

Bytes Written:        43200000713
Files Written:        20

### DIFF
--- a/io/encryption/writer.go
+++ b/io/encryption/writer.go
@@ -26,6 +26,7 @@ import (
 type writer struct {
 	writer io.WriteCloser
 	stream cipher.Stream
+	buf    []byte
 }
 
 // NewWriter returns a new encryption Writer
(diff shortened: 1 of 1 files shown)


io/encryption/writer.go
```diff
@@ -63,7 +64,10 @@ func NewWriter(w io.WriteCloser, key []byte) (io.WriteCloser, error) {
 
 // Write writes the data to the writer and encrypts it.
 func (ew *writer) Write(p []byte) (int, error) {
-	encrypted := make([]byte, len(p))
+	if cap(ew.buf) < len(p) {
+		ew.buf = make([]byte, len(p))
```
filkeith, line 68: consider resetting buffer if it grows too much, smth like

if cap(ew.buf) > maxCap {
     ew.buf = nil
}

PR #434, opened 2026-04-13
Fieldalign token struct

korotkov-aerospike:
Fieldalign says: 16 bytes saved: struct with 48 pointer bytes could be 32

No measurable improvement, but let's keep it just for piece in mind. We use token a lot.


filkeith (approved):
lgtm

PR #437, opened 2026-04-19
BRKS-178 Reader refactor

korotkov-aerospike:


io/aerospike/record_reader.go
```diff
@@ -119,205 +124,258 @@ func newSingleRecordReader(
...
 
-		r.config.rpsCollector.Increment()
+// ensureActiveScan starts a new scan if none is active, respecting context cancellation.
+func (r *singleRecordReader) ensureActiveScan(ctx context.Context) error {
+	r.mu.Lock()
+	hasActive := r.active != nil
+	r.mu.Unlock()
 
-		return recToken, nil
+	if hasActive {
```
filkeith: you changing var with mutex, but reading it without mutex. It would be better to read it with mutex too

io/aerospike/record_reader.go
```diff
@@ -119,205 +124,258 @@ func newSingleRecordReader(
...
+		return nil
 	}
-}
 
-// Close no-op operation to satisfy pipe.Reader interface.
-func (r *singleRecordReader) Close() {
+	// Check both the caller's context and our own internal context.
+	select {
+	case <-ctx.Done():
+		r.cancel()
```
filkeith, line 174: need to test it with cancellation, how backup interruption will work

korotkov-aerospike:
1. Need to update paginated reader too
2. Check mutexes and logic in general
3. Verify cancelation with unit tests

korotkov-aerospike:
tests passed https://automation.ecoeng.dev/jenkins/job/abs/job/multi-node-local-env/job/master/24/

pipe/chains.go
```diff
@@ -76,6 +80,10 @@ func newReaderRoutine[T models.TokenConstraint](r Reader[T], p Processor[T], out
 		defer close(output)
 
 		for {
+			if err := ctx.Err(); err != nil {
```
filkeith, line 83: Hope this won't affect restore speed

filkeith (approved):
Logic beacme complex...

PR #442, opened 2026-05-03
BKRS-182 Restore decoder optimization

korotkov-aerospike:
### 1. Network I/O System Calls (The 1MB Buffer Fix)
**The Problem:**
The S3 network stream was wrapped in a default `bufio.Reader` (4KB buffer). As the parser decoded the backup line-by-line, the tiny buffer size forced the runtime to issue over hundreds of thousands of tiny `read()` system calls over the network socket. This resulted in `syscall.rawsyscalln` consuming **~39% of total CPU time**.

**The Fix:**
We increased the buffer size at the decoder's entry point (`io_encoding.go`) to **1 Megabyte** using `bufio.NewReaderSize(src, 1024*1024)`. 

**Impact:**
Because `newCountingReader` does not double-wrap an existing adequately-sized `bufio.Reader`, this successfully batched the network IO, dropping `syscall.rawsyscalln` overhead massively and saving over **10 seconds** of raw wall time.

### 2. Eliminating the Memory Pool Leak
**The Problem:**
The decoder utilized three separate memory pools (`bigBufPool`, `smallBufPool`, `base64BufferPool`). A bug in `_readBlockDecodeBase64` caused buffers allocated from the `bigBufPool` to be returned to the `smallBufPool`. 
This starved the `bigBufPool` (forcing constant, heavy `make([]byte)` allocations) and flooded the `smallBufPool` with massive buffers, resulting in tremendous Garbage Collection pressure.

**The Fix:**
We unified all three pools into a single `bufPool`. Slices are fetched from the pool, and if they are too small, they are dynamically re-allocated `make([]byte, size)`. When they are returned `bufPool.Put(buf[:0])`, the capacity is retained.

**Impact:**
Memory allocation and garbage collection CPU time (`runtime.memclrNoHeapPointers`) dropped significantly, yielding roughly a **15% reduction in absolute GC CPU time**.

### 3. Fast-Path Byte Parsing (`_readUntilByte`)
**The Problem:**
The `_readUntilAny` function was responsible for parsing tokens character-by-character. For every byte, it called `slices.Contains(delims, b)`. Even though this loop was small, `slices.Contains` introduces a generic loop that prevents compiler optimizations inside a hot loop executed millions of times.

**The Fix:**
We created a dedicated `_readUntilByte` function for the 99% use-case where we are searching for a *single* delimiter (e.g., `\n` or ` `). This removes the length check entirely and compares the byte natively. For multi-delimiter searches, we migrated `_readUntilAny` to use the Assembly-optimized `bytes.IndexByte()`.

**Impact:**
The CPU time spent inside `_readUntilAny` dropped from **1.64% cumulative** down to a nearly invisible **0.26%**.

---

### Conclusion
By batching network system calls, stopping memory pool leaks, and shaving instructions out of the byte-parser loop, the overall end-to-end restore speed increased by **over 34%**. The restore process is now almost completely constrained by raw Network I/O and standard Go slice memory allocations, representing a highly optimized Go processing pipeline.

One more test, with different data

 Comparison Summary:

| Metric            | Main    | Optimized | Improvement |
| ----------------- | ------- | --------- | ----------- |
| Total CPU samples | 155.46s | 124.54s   | -19.9%      |
| CPU %             | 308.06% | 257.36%   | -16.5%      |
| Wall time         | 53.28s  | 51.16s    | -4.0%       |

### Decoder function comparison (cumulative time)

| Function                | Main           | Optimized      | Change  |
| ----------------------- | -------------- | -------------- | ------- |
| countingReader.ReadByte | 37.89s (24.4%) | 9.92s (8.0%)   | -73.8%  |
| _readUntilAny           | 24.02s (15.5%) | —              | removed |
| _readUntil              | 13.86s (8.9%)  | —              | removed |
| readUntilByteEscaped    | —              | 5.62s (4.5%)   | new     |
| readUntilByte           | —              | 3.39s (2.7%)   | new     |
| readRecord              | 49.75s (32.0%) | 27.50s (22.1%) | -44.7%  |
| readSignedInt           | 5.73s (3.7%)   | 2.62s (2.1%)   | -54.3%  |

  The optimizations are confirmed and significant. The decoder is now much more efficient - countingReader.ReadByte went from 24% of CPU time to just 8%,
  and overall decoder time (readRecord) dropped by ~45%.


filkeith (dismissed):
lgtm

korotkov-aerospike:
tests pass https://automation.ecoeng.dev/jenkins/job/abs/job/multi-node-local-env/job/master/26/

filkeith (approved):
We need to test it with all data types

PR #452, opened 2026-05-17
BKRS-190 Add estimaets printing as shared lib

filkeith:
- moved estimates printing from CLI to backup-go lib

### DIFF
--- a/pkg/estimates/estimates.go
+++ b/pkg/estimates/estimates.go
@@ -128,17 +128,17 @@ func printEstimate(
 	getMetrics func() *models.Metrics,
 	logger *slog.Logger,
 ) (float64, error) {
-	percentage := done / total
-	estimatedEndTime := calculateEstimatedEndTime(startTime, percentage)
+	ratio := done / total
+	estimatedEndTime := calculateEstimatedEndTime(startTime, ratio)
 
 	switch {
-	case percentage >= 1:
+	case ratio >= 1:
 		// Exit after 100%.
 		return 0, errBrake
-	case percentage*100 < 1:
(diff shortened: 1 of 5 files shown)


pkg/estimates/estimates.go
```diff
@@ -0,0 +1,221 @@
...
+
+func printEstimate(
+	startTime time.Time,
+	done,
+	total,
+	previousVal float64,
+	getMetrics func() *models.Metrics,
+	logger *slog.Logger,
+) (float64, error) {
+	percentage := done / total
```
korotkov-aerospike: it's not percentage, it's ratio (0 - 1). It should be float, but done and total are integers

pkg/estimates/estimates.go
```diff
@@ -0,0 +1,221 @@
...
+	estimatedEndTime := calculateEstimatedEndTime(startTime, percentage)
+
+	switch {
+	case percentage >= 1:
```
korotkov-aerospike: don't break on 100%. Break when ctx is finsihed.

filkeith: Why? 100% means we won't print anything more, so exit

pkg/estimates/estimates.go
filkeith: renamed

pkg/estimates/estimates.go
filkeith: There is conterxt check PrintRestoreEstimate, so on the next tick it will be cheked again. So no need to check it here. Everything is correct

korotkov-aerospike (approved):

PR #464, opened 2026-06-07
SERVER-733 Add info command to show server backup progress

filkeith:


pkg/asinfo/client.go
```diff
@@ -1043,6 +1048,95 @@ func (ic *Client) PrepareServerRestore(ctx context.Context, jobID, namespace str
...
+
+			if trID != firstTrID {
+				return fmt.Errorf("backup is starting on node %s, but not all nodes have the same trid: %s",
+					node.GetName(), node.GetName())
+			}
+
+			total += one
+		}
+
+		result = math.Min(100, 100*total/4096)
```
korotkov-aerospike: I would really recommend to have 0..1 as internal representation and only convert to percentages for io to user

pkg/asinfo/client.go
```diff
@@ -1474,31 +1584,49 @@ type infoMap map[string]string
...
+		return 0, true, fmt.Errorf("failed to parse %s=%q: %w", key, val, err)
 	}
 
-	return 0, false, nil
+	return v, true, nil
+}
+
+// parseFloat64 returns the parsed float64 value from the map for the given key if found.
+// ok = true if the key was found.
+func (m infoMap) parseFloat64(key string) (result float64, ok bool, err error) {
```
korotkov-aerospike: Maybe extract parseXXX functions into separate pacakage?

pkg/asinfo/client.go
```diff
@@ -1651,3 +1779,45 @@ func bitMapToIntSlice(b []bool) []int {
 
 	return result
 }
+
+// filterBackupsSortedByTimeSinceDone filters backup jobs from the given list
+// and sorts them by time-since-done in ascending order.
+func filterBackupsSortedByTimeSinceDone(jobs []infoMap) ([]infoMap, error) {
```
korotkov-aerospike: you accept jobs, modify it in-place and return.
That might be confusing. Either return new slice (that's what I would do), or just filter/sort in place and return only error.

pkg/asinfo/client.go
filkeith: changed

pkg/asinfo/client.go
filkeith: moved

pkg/asinfo/client.go
filkeith: fixed

pkg/asinfo/utils.go
```diff
@@ -0,0 +1,489 @@
...
+			result = append(result, i)
+		}
+	}
+
+	return result
+}
+
+// filterBackupsSortedByTimeSinceDone filters backup jobs from the given list
+// and sorts them by time-since-done in ascending order.
+func filterBackupsSortedByTimeSinceDone(jobs1 []m.InfoMap) ([]m.InfoMap, error) {
```
korotkov-aerospike: jobs1 -> jobs

client.go
```diff
@@ -23,6 +23,7 @@ import (
 	"github.com/aerospike/backup-go/internal/scanlimiter"
 	"github.com/aerospike/backup-go/models"
 	"github.com/aerospike/backup-go/pkg/asinfo"
+	models2 "github.com/aerospike/backup-go/pkg/asinfo/models"
```
korotkov-aerospike, line 26: "models2" ?

filkeith: opps,  temporary naming

korotkov-aerospike (approved):

PR #488, opened 2026-08-05
BKRS-271 Harden local backup path handling

korotkov-aerospike:
## Summary
This PR hardens local and cloud storage path handling and file lifecycle behavior to reduce traversal, symlink, and TOCTOU risks while preserving expected backup/restore flows.

### Local storage hardening (`io/storage/local`)
- Moved local reader/writer directory operations to `os.Root`-scoped access for safer open/list/remove behavior.
- Replaced stat-then-act patterns with direct operations to reduce TOCTOU race windows.
- Added explicit handling for file-vs-directory removal (`isNotDir`) and clarified behavior with comments in non-obvious branches.
- Enforced restrictive creation modes (`0700` for directories, `0600` for files) and `O_TRUNC` on writes.
- Added local filename validation in `io/storage/local/path_validator.go` and validated names before touching filesystem in `Writer.NewWriter`.

### Cloud object key hardening (`io/storage/common`)
- Added `ValidateObjectKey` and integrated it into `common.GetFullPath()`.
- Rejects unsafe key patterns (`NUL`, leading `/`, `..` path segments) while allowing nested object keys with `/`.
- Updated Azure/GCP tests to use valid filename inputs passed to writer APIs.

### Backup continuation resilience
- Keeps state-writing synchronization fix that avoids partial state reads/EOF during continuation shutdown.

## Test plan
- `go test ./io/storage/common/... ./io/storage/local/... -count=1`
- `go test ./io/storage/local -count=1`
- `gci write . && golangci-lint run ./io/storage/common/... ./io/storage/local/...`
- CI: run full `make coverage` suite in GitHub Actions


io/storage/local/writer.go
```diff
@@ -104,7 +104,7 @@ func createDirIfNotExist(path string, isDir bool) error {
 	case err == nil:
 		// ok.
 	case os.IsNotExist(err):
-		if err = os.MkdirAll(path, os.ModePerm); err != nil {
+		if err = os.MkdirAll(path, 0o700); err != nil {
```
filkeith: os.ModePerm is 0777

io/storage/common/path_validator.go
```diff
@@ -0,0 +1,55 @@
...
+
+	if strings.ContainsAny(filename, `/\`) {
+		return fmt.Errorf("filename must not contain path separators ('/' or '\\')")
+	}
+
+	if strings.ContainsRune(filename, '\x00') {
+		return fmt.Errorf("filename must not contain NUL bytes")
+	}
+
+	if filepath.IsAbs(filename) {
```
filkeith, line 46: this one is already covered with `strings.ContainsAny(filename, `/\`)`.

io/storage/common/writer.go
```diff
@@ -21,6 +21,10 @@ import (
 
 // GetFullPath returns full path for file or directory, according to params.
 func GetFullPath(prefix, filename string, pathList []string, isDir bool) (string, error) {
+	if err := ValidateFilename(filename); err != nil {
```
filkeith, line 24: we should check only if `isDir = false` ? Can't we receive path here as filename?

io/storage/local/writer.go
```diff
@@ -138,75 +129,71 @@ func (w *Writer) Remove(ctx context.Context, targetPath string) error {
...
+			return nil
 		}
 
-		return nil
+		return fmt.Errorf("failed to open %s: %w", targetPath, err)
+	}
+
+	defer root.Close()
+
+	f, err := root.Open(".")
```
filkeith, line 161: Can you add some comments to the new code, as not everything is obvious. I'm afraid we forgot in half a year what it is.

filkeith (changes requested):

korotkov-aerospike:
Autiomation tests on Jenkins passed

filkeith (approved):
LGTM (if regression tests passes)

PR #495, opened 2026-08-09
BKRS-326 Remove xdr related code

filkeith:


korotkov-aerospike (approved):
Wow, I really like how many dead code is cleaned.

PR #496, opened 2026-08-12
BKRS-334 Add support of set index

filkeith:


models/record.go
```diff
@@ -44,6 +44,7 @@ const (
 	StringSIDataType      SIPathBinType = 'S'
 	GEO2DSphereSIDataType SIPathBinType = 'G'
 	BlobSIDataType        SIPathBinType = 'B'
+	EmptySIDataType       SIPathBinType = 'E' // For index without bins.
```
korotkov-aerospike: Please, add more explanation, what is this index.

handler_backup.go
```diff
@@ -641,3 +641,20 @@ func (bh *BackupHandler) cleanup() {
 
 	bh.pl.Swap(nil)
 }
+
+func getSIndexInfo(ctx context.Context, infoClient ClusterInfo, namespace string) (*models.SIndexInfo, error) {
```
korotkov-aerospike: getSIndexInfo calls GetSIndexes twice — HasExpressionSIndex and HasSetSIndex. Works, but redundant network round-trips.

korotkov-aerospike:
Did you run e2e tests? I think it's worth adding: backup index, then delete it and restore.

models/record.go
filkeith: added

handler_backup.go
filkeith: replaced with 1 info command

filkeith:
> Did you run e2e tests? I think it's worth adding: backup index, then delete it and restore.

@korotkov-aerospike Yes, I did it manually. Restore to a clean DB and check that the new index exists

korotkov-aerospike (approved):

PR #498, opened 2026-08-16
Optimize ASB record encoding

korotkov-aerospike:
## Summary

This PR replaces the production ASB record encoder with a faster append-based implementation and switches `EncodeToken` to a reused `[]byte` API. The previous `bytes.Buffer` encoder is kept in `internal/legacy_encoder` only for byte-parity tests and side-by-side benchmarks.

### Why

Record encoding is on the backup hot path. The old encoder layered many `bytes.Buffer.Write` helpers. The new encoder builds each record with a flat `[]byte` append chain, single-pass escaping, namespace/set line caching, and precomputed generation / bin-count / never-expire lines.

#### 1. `EncodeToken` API: `*bytes.Buffer` → `[]byte`

```go
EncodeToken(token *models.Token, dst []byte) ([]byte, error)
```

Callers reuse a backing slice:

```go
buf = buf[:0]
buf, err = encoder.EncodeToken(token, buf)
```

Updated call sites: `token_writers.go`, `handler_backup.go`, tests, and `mocks/Encoder_mock.go`.

#### 2. Production encoder in `encode.go`

- Flat append chain instead of nested buffer writes
- Single-pass escaping via `appendEscapedDirect`
- Namespace/set `recentLine` cache for sequential records
- Precomputed generation, bin-count, and `+ t 0` lines

UDF, SIndex, and file headers use the same append path.

#### 3. `internal/legacy_encoder` for verification

Not used in production. Parity tests: `TestRecordEncoderMatchesLegacy`, `TestEncodeTokenRecordMatchesLegacy`. Comparison bench: `BenchmarkRecordEncoderWorkloads`.

```sh
go test -bench=BenchmarkRecordEncoderWorkloads -benchmem -run=^$ ./io/encoding/asb/internal/legacy_encoder
go test -bench=BenchmarkEncodeRecord -benchmem -run=^$ ./io/encoding/asb ./io/encoding/asb/internal/legacy_encoder
```

### Benchmarks (Apple M-series, darwin/arm64, `-count=3`, warm reused buffer)

All implementations: **0 B/op, 0 allocs/op**.

`BenchmarkRecordEncoderWorkloads` (median of 3):

| Workload | Legacy | EncodeToken | Speedup |
| --- | ---: | ---: | ---: |
| small_scalar | 248 ns/op | 110 ns/op | **2.3×** |
| escaped_metadata | 284 ns/op | 126 ns/op | **2.3×** |
| long_escaped_names | 336 ns/op | 175 ns/op | **1.9×** |
| medium_all_types | 1041 ns/op | 599 ns/op | **1.7×** |
| medium_all_types_compact | 977 ns/op | 552 ns/op | **1.8×** |
| many_int_bins | 2388 ns/op | 1380 ns/op | **1.7×** |
| large_compact_64k | 1014 ns/op | 952 ns/op | **1.1×** |
| large_base64_64k | 5255 ns/op | 5116 ns/op | **1.03×** |

`BenchmarkEncodeRecord` (same mixed 10-bin record, including a 1 KiB blob):

| Implementation | ns/op | allocs |
| --- | ---: | ---: |
| Legacy (`bytes.Buffer`) | ~916–975 | 0 |
| New (`EncodeToken`) | ~502–512 | 0 |

`BenchmarkEncodeTokenManySets` (256 rotating set names, cache misses): **128 ns/op, 0 B/op, 0 allocs/op**.

Large base64 payloads stay encode-bound; metadata-heavy records see the largest win.

### Review follow-up

- `TestEncodeTokenManySets` encodes 128 distinct sets on one encoder and matches a fresh encoder
- `TestEncodeTokenReusedBufferDoesNotGrowUnbounded` encodes 2048 records into one slice; capacity stays under 4 KiB
- Memory: `-benchmem` reports 0 allocations on the hot path when the caller reuses `dst`

### Risks / notes

- Multi-bin map iteration order is still non-deterministic (parity tests sort bin lines)
- Namespace/set cache is per-encoder and updates when the value changes
- Partial output on error is preserved (`TestRecordEncoderPreservesPartialOutputOnError`)

## Test plan

- [x] `gci write . && golangci-lint run ./io/encoding/asb/...`
- [x] `go test ./io/encoding/asb/... -count=1`
- [x] Legacy vs new parity tests
- [x] Local benchmarks: `BenchmarkRecordEncoderWorkloads`, `BenchmarkEncodeRecord`
- [ ] Integration tests (live Aerospike)


filkeith:
It is not good to name the PR with another task BKRS-271 is about path traversal in storage operations

io/encoding/asb/encode.go
```diff
@@ -105,821 +116,418 @@ func (e *Encoder[T]) headerVersion(isRecords bool) string {
...
+	dst = append(dst, globalSection...)
+	dst = append(dst, space...)
+
+	if sindex.Expression != "" {
+		dst = append(dst, globalSIndexExpression...)
+	} else {
+		dst = append(dst, globalSIndex...)
+	}
+
+	dst = append(dst, space...)
```
filkeith, line 509: I'm afraid that this ammount of appends will make GC crazy

io/encoding/asb/encode.go
```diff
@@ -27,11 +27,26 @@ import (
...
+	}
+}
+
 // Encoder contains logic for encoding backup data into the .asb format.
 // This is a stateful object that must be created for every backup operation.
 type Encoder[T models.TokenConstraint] struct {
-	config *EncoderConfig
-
+	config           *EncoderConfig
+	recordNamespace  recentLine
```
filkeith, line 53: namespace is already 1 for backup, we can't backup 2 namespaces at once

filkeith (changes requested):
need to do more tests with a bunch of different sets, also measure memory consumption

io/encoding/asb/encode.go
korotkov-aerospike: no, it's all the same []byte.
It will be allocated for the first record, then it will reuse same buffer over and over

io/encoding/asb/encode.go
korotkov-aerospike: ok, then we will have no cache misses. But I would keep it for the correctness.

io/encoding/asb/encode.go
korotkov-aerospike: check the benchmark. We have 0 allocations.

filkeith (approved):
lgtm

PR #501, opened 2026-08-17
BKRS-334 Add support of set index

filkeith:


io/encoding/asb/config.go
```diff
@@ -23,19 +25,31 @@ type EncoderConfig struct {
...
+	HasSetSIndex bool
 }
 
 // NewEncoderConfig returns a new encoder EncoderConfig.
-func NewEncoderConfig(namespace string, compact, hasExpressionSIndex bool) *EncoderConfig {
-	return &EncoderConfig{
-		Namespace:           namespace,
-		Compact:             compact,
-		HasExpressionSIndex: hasExpressionSIndex,
+func NewEncoderConfig(namespace string, compact bool, sIndexInfo *models.SIndexInfo) *EncoderConfig {
```
korotkov-aerospike: maybe pass sIndexInfo as a struct, not ref? 
then we will not need to check for if sIndexInfo != nil {

korotkov-aerospike (approved):

io/encoding/asb/config.go
filkeith: fixed

PR #502, opened 2026-08-17
BKRS-334 Skip unknown index types

filkeith:
- skip unknown index types

### DIFF
--- a/pkg/asinfo/utils.go
+++ b/pkg/asinfo/utils.go
@@ -145,12 +145,12 @@ func (ic *Client) parseSIndexes(sindexListInfoResp string, noWarn bool) ([]*mode
 		return nil, fmt.Errorf("failed to parse sindex response: %w", err)
 	}
 
-	// No sindexes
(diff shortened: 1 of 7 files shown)


pkg/asinfo/utils.go
```diff
@@ -136,31 +139,54 @@ func parseAerospikeVersion(versionStr string) (m.AerospikeVersion, error) {
...
+			continue
 		}
 
-		sindexes[i] = sindex
+		sindex, err := parseSIndex(sindexMap)
+
+		switch {
+		case err == nil:
+			sindexes = append(sindexes, sindex)
+		case errors.Is(err, ErrInvalidSIndexType):
```
korotkov-aerospike, line 166: I am not sure it's a good idea to pass logger just for this case.
Let's discuss tomorrow, but I would suggest to add one more return value - unparsed indexes. But let's talk tomorrow

korotkov-aerospike (approved):

PR #520, opened 2026-09-06
BKRS-386 Add nilaway linter

korotkov-aerospike:
## Summary

Adds Uber's [NilAway](https://github.com/uber-go/nilaway) as a production-only CI gate (`make nilaway`, tests and generated mocks excluded) and fixes the findings that were real or that the analyzer could not prove from local interfaces.

**CI**
- Pin NilAway and run it on every push/PR.
- Document `make nilaway` in `CONTRIBUTING.md`.

**Public API**
- `Restore` rejects a nil streaming reader.
- `Backup` requires a reader only when **continuing** from a state file (first-run state backup does not read it).
- `Estimate` rejects `StateFile` (estimate never uses state).
- `NewState` no longer returns `(nil, nil)`; continue-from-state requires a reader.

**Storage / encode / scan**
- Treat a nil S3 `UploadPart` response and nil GCS object attrs as errors.
- Reject a typed-nil `*RawBlobValue` when encoding ASB.
- Reject an empty primary-partition list from the cluster.
- Reject a nil Aerospike scan `*Result` before reading `Err`.

**ASB decode**
- Replace the Peek-based `readUntilAny` with `readUntilWhitespace` (metadata tokens).
- `readUntilByte` keeps `ReadSlice` for speed, loops on `ErrBufferFull` so tokens can span the 1 MB `bufio` fill (records larger than 1 MB, up to at least 8 MB, decode correctly), unread the delimiter only when it was actually consumed, and updates line/column the same way `ReadByte` does.

Empty `PartitionFilters` is unchanged: it still means all partitions, except paginated backup (`PageSize > 0`) which already required filters.

## Test plan

- [ ] `make nilaway` is green on this branch and red if the new guards are reverted.
- [ ] `make test-unit` / CI unit tests (race) pass.
- [ ] Restore with a nil reader returns a clear error.
- [ ] Estimate with `StateFile` set returns an error.
- [ ] First-run backup with `StateFile` set still works with a nil reader; continue-from-state without a reader fails.
- [ ] ASB decode: `readUntilByte` tests (unread + tracker, token spanning small buffer fills, `skipToNextLine`).
- [ ] Spot-check restore of a backup that includes a large record (multi-MB bin) if an 8 MB fixture is available.


filkeith (approved):
lgtm but you need to update the branch

filkeith (approved):

PR #522, opened 2026-09-07
BKRS-387 Add all data types test

filkeith:
- added test

### DIFF
--- a/backup_restore_integration_test.go
+++ b/backup_restore_integration_test.go
@@ -22,11 +22,18 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"math"
+	"math/rand/v2"
 	"os"
 	"path"
+	"reflect"
+	"slices"
+	"strconv"
+	"strings"
 	"sync"
 	"testing"
 	"time"
+	"unicode/utf8"
 
 	a "github.com/aerospike/aerospike-client-go/v8"
 	"github.com/aerospike/backup-go/io/encoding/asb"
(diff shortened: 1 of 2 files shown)


backup_restore_integration_test.go
```diff
@@ -52,6 +59,182 @@ const (
...
+	dtClassScalar = dtSizeClass{name: "scalar", size: dtSizeScalar, records: dtRecordsDefault}
+	dtClassSmall  = dtSizeClass{name: "1536b", size: dtSizeSmall, records: dtRecordsDefault}
+	dtClassMedium = dtSizeClass{name: "100kb", size: dtSizeMedium, records: dtRecordsDefault}
+	dtClassLarge  = dtSizeClass{name: "1mb", size: dtSizeLarge, records: dtRecordsLarge}
+	dtClassHuge   = dtSizeClass{name: "7mb", size: dtSizeHuge, records: dtRecordsHuge}
+)
+
+// dtWriter populates one test set and reports the bins the database is expected
+// to hold, keyed by record digest.
+type dtWriter interface {
```
korotkov-aerospike, line 227: d**b**Writer?

korotkov-aerospike: or it's dataWriter?

korotkov-aerospike (approved):

backup_restore_integration_test.go
filkeith: It exists because one type behaves differently from the rest. dtGenWriter builds bins on the client, writes them with Put, and returns exactly what it generated. dtHLLWriter can't do that — an HLL particle is built by the server — so it writes via Operate(HLLAddOp) and reads the record back to learn what the value actually is. Same output shape, two different paths; dtRunCase doesn't care which one it got.

